PR #39, opened 2016-07-27
server: makes Conn an interface

### LINKED ISSUE #33: Find a way to allow third-party packages to extend server.Conn
* `ID` extension needs to store IDs for each conn
* `ENABLE` extension needs to keep a list of enabled extensions for each conn

### LINKED ISSUE #42: Add an extension interface
Add a `Server.Enable(extension Extension)` method.

```go
type Extension interface {
    Capabilities(state common.ConnState) []string
    Commands() map[string]HandlerFactory
}
```

emersion:
Fixes #33
Fixes #42

This breaks compatibility with all extensions.


emersion:
Closing in favor of #44

PR #41, opened 2016-08-07
common: makes Writer an interface

emersion:
External packages should not have access to Writer.

Breaks compatibility with dates and extensions.


emersion:
Closed in favor of #45

PR #50, opened 2016-08-17
When searching fall back to US-ASCII if BADCHARSET is received.

### LINKED ISSUE #48: Exchange 2013: The specified charset is not supported.
It looks like ms exchange does not like searching with UTF-8. I have found a few pages out there talking about it. 

1. https://social.technet.microsoft.com/Forums/en-US/00bb2ec5-9951-40b4-8ec3-97e5a8f1a59f/imap-search-with-utf8?forum=exchangesvrsecuremessaging

2. MatterMail (which wraps mxk/go-imap) is trying UTF-8 first and then falling back to US-ASCII. https://github.com/rodrigocorsi2/mattermail/issues/7

chris-garrett:
Fixes #48


emersion:
LGTM, thanks!

PR #54, opened 2016-09-08
make the Store flags case insensitive

zhj4478:


emersion:
Build is failing. You didn't test that your code was working.

Additionally, there are many other places where lists fo flags are created. Also, this happens not only in parse functions.

zhj4478:
the build failed maybe because "github.com/emersion/go-imap" have been removed by my purifier tool. 
The code works fine on my local env.
if i am not wrong from the rfc those flags should be caseinsensitive.
since you konw where the issue is, I am not familiar with this code, if you have time it is better you make this change. so ignore this PR.
@emersion

emersion:
Fixed by https://github.com/emersion/go-imap/commit/caf09eac64fc41c337be2fb54040a5d55ff67a99

PR #59, opened 2016-09-30
ENVELOPE should use the RFC1123Z time format.

hartshorne:
RFC 3501 uses this format in its sample connection:

  https://tools.ietf.org/html/rfc3501#section-8

Seems like this format is inspired by RFC2822's specification:

  https://tools.ietf.org/html/rfc2822#section-3.3

FastMail and Gmail also the RFC1123Z format for ENVELOPE.

### DIFF
--- a/client/cmd_selected_test.go
+++ b/client/cmd_selected_test.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"net"
 	"testing"
+	"time"
 
 	"github.com/emersion/go-imap"
 	"github.com/emersion/go-imap/client"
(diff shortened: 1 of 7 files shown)


hartshorne:
The tests fail; not sure what you want to do here. Seems kind of silly to have ParseDate and ParseDateTime functions that just wrap time.Parse.

emersion:
Yeah, a better solution would probably to add `imap.DateTimeFormat` and `imap.DateFormat`, what do you think?

hartshorne:
Sure, but what do you want to do in `func (w *Writer) writeField(field interface{}) error`? INTERNALDATE should probably be in 2-Jan-2006 15:04:05 -0700 format, and ENVELOPE should be in time.RFC1123Z format.

That said, RFC 2822 seems to think the day-of-week is optional, so safe to always write as 2-Jan-2006 15:04:05 -0700?

PR #60, opened 2016-09-30
ENVELOPE should be able to parse the RFC1123Z time format.

hartshorne:
RFC 3501 uses this format in its sample connection:

  https://tools.ietf.org/html/rfc3501#section-8

Seems like this format is inspired by RFC2822's specification:

  https://tools.ietf.org/html/rfc2822#section-3.3

FastMail and Gmail also the RFC1123Z format for ENVELOPE.


hartshorne:
search.go is broken

search.go
```diff
@@ -132,19 +132,19 @@ func (c *SearchCriteria) Parse(fields []interface{}) error {
...
-			c.SentSince, _ = ParseDate(fields[i])
+			c.SentSince, _ = time.Parse(DateFormat, fields[i].(string))
 		case "SINCE":
 			i++
-			c.Since, _ = ParseDate(fields[i])
+			c.Since, _ = time.Parse(DateFormat, fields[i].(string))
 		case "SMALLER":
 			i++
-			c.Smaller, _ = ParseNumber(fields[i])
+			c.Smaller, _ = ParseNumber(fields[i].(string))
```
hartshorne: whoops

search.go
```diff
@@ -132,19 +132,19 @@ func (c *SearchCriteria) Parse(fields []interface{}) error {
 			c.Seen = true
 		case "SENTBEFORE":
 			i++
-			c.SentBefore, _ = ParseDate(fields[i])
+			c.SentBefore, _ = time.Parse(DateFormat, fields[i].(string))
```
hartshorne: this introduces a potential crash, need to check that the assertion worked

write.go
```diff
@@ -177,8 +169,10 @@ func (w *Writer) writeField(field interface{}) error {
 		return w.writeLiteral(field)
 	case []interface{}:
 		return w.writeList(field)
+	case *timeLayout:
```
hartshorne: seem reasonable?

search.go
```diff
@@ -48,6 +48,15 @@ type SearchCriteria struct {
 	Unseen     bool
 }
 
+func maybeString(mystery interface{}) string {
```
hartshorne: t.Format returns a 0'd date when given an invalid (or empty) string

    t, _ := time.Parse(time.RFC822, "")
    println(t.Format(time.RFC822)) // "01 Jan 01 00:00 UTC"

emersion: Seems reasonable in this particular case because errors are ignored anyway when parsing fields.

write.go
emersion: 1. Extensions in other packages should be able to specify in which format they want dates to be formatted.
2. Maybe something like `Quoted` would be better? The idea is to add:

   ```go
   type Date time.Time
   type DateTime time.Time
   ```

   And use the corresponding format in this switch. Another solution would be to add some exported functions to create `timeLayout`s.

hartshorne: How about something like this? Fewer conditionals, can just thread whatever you need through.

	// A time.Time with a specific layout
	type TimeLayout struct {
		time.Time
		layout string
	}

	func (d *TimeLayout) SetLayout(layout string) {
		d.layout = layout
	}

	func (d *TimeLayout) Parse(value string) (time.Time, error) {
		return time.Parse(d.layout, value)
	}

	func (d *TimeLayout) FormatLayout() string {
		return d.Format(d.layout)
	}

emersion: What I don't like with it is that it lets you specify ANY time layout...

message.go
```diff
@@ -513,14 +513,78 @@ type Envelope struct {
...
+	"Mon, 02 Jan 06 15:04:05 MST",
+	"Mon, 02 Jan 06 15:04:05 -0700 (MST)",
+	"Mon, 02 Jan 06 15:04 -0700",
+	"Mon, 02 Jan 06 15:04 MST",
+	"Mon, 02 Jan 06 15:04 -0700 (MST)",
+}
+
+// Try parsing the date based on the layouts defined in RFC 5322, section 3.3.
+// Inspired by https://github.com/golang/go/blob/master/src/net/mail/message.go
+func (e *Envelope) ParseDate(maybeDate string) (time.Time, error) {
```
hartshorne: maybe organize all the new Date/Time stuff into a new file, date.go? probably want to be able to parse all the IMAP-style date/times as well.

emersion: Yeah, that would be better. Also, I'm not sure we want to export this parse function as it's not directly related to IMAP (it should be defined in `net/mail` in a perfect world).

I can help if you want (I just don't want to do something you don't want me to :P ).

hartshorne:
FastMail seems to send an INTERNALDATE with a leading space. Let me double check.

hartshorne:
Yeah, FastMail sends a leading space:

  SS0SbQ FETCH 1:* (INTERNALDATE)
  * 1 FETCH (INTERNALDATE " 1-Jun-2016 14:42:13 -0400")
  * 2 FETCH (INTERNALDATE " 1-Jun-2016 14:47:37 -0400")
  * 3 FETCH (INTERNALDATE " 2-Jun-2016 01:28:37 -0400")
  * 4 FETCH (INTERNALDATE "22-Jul-2016 13:36:40 -0400")

That may be part of the RFC 1730 date spec (page 55):

   date            ::= date_text / <"> date_text <">

   date_day        ::= 1*2digit
                       ;; Day of month

   date_day_fixed  ::= (SPACE digit) / 2digit
                       ;; Fixed-format version of date_day

search_test.go
```diff
@@ -33,10 +33,10 @@ var searchCriteriaTests = []struct {
 			"KEYWORD", "cc",
 			"LARGER", "4242",
 			"NEW",
-			"NOT", []interface{}{"OLD", "ON", "5-Nov-1984"},
-			"OR", []interface{}{"RECENT", "SENTON", "21-Nov-1997"}, []interface{}{"SEEN", "SENTBEFORE", "5-Nov-1984"},
+			"NOT", []interface{}{"OLD", "ON", " 5-Nov-1984"},
+			"OR", []interface{}{"RECENT", "SENTON", "21-Nov-1997"}, []interface{}{"SEEN", "SENTBEFORE", " 5-Nov-1984"},
 			"SENTSINCE", "21-Nov-1997",
-			"SINCE", "5-Nov-1984",
+			"SINCE", " 5-Nov-1984",
```
hartshorne: you sure this is what you want the server to do?

hartshorne: Cyrus pads with leading spaces:
https://github.com/cyrusimap/cyrus-imapd/blob/1cb805a3bffbdf829df0964f3b802cdc917e76db/lib/times.c#L543

Dovecot pads with leading zeros:
https://github.com/dovecot/core/blob/4fbd5c5e113078e72f29465ccc96d44955ceadc2/src/lib-imap/imap-date.c#L166

FastMail uses Cyrus:
https://www.fastmail.com/help/technical/architecture.html

date.go
```diff
@@ -0,0 +1,88 @@
...
+//   https://github.com/dovecot/core/blob/4fbd5c5e113078e72f29465ccc96d44955ceadc2/src/lib-imap/imap-date.c#L166
+// Cyrus adds a leading space to dates:
+//   https://github.com/cyrusimap/cyrus-imapd/blob/1cb805a3bffbdf829df0964f3b802cdc917e76db/lib/times.c#L543
+const (
+	// Described in RFC 1730 on page 55.
+	DateLayout = "_2-Jan-2006"
+	// Described in RFC 1730 on page 55.
+	DateTimeLayout = "_2-Jan-2006 15:04:05 -0700"
+	// Described in RFC 5322 section 3.3.
+	messageDateTimeLayout = "Mon, 02 Jan 2006 15:04:05 -0700"
```
emersion: This layout is referred to in RFC 3501 as `env-date` (but is not defined in this RFC, it's just a `nstring`), maybe it's better to name it `envelopeDateTimeLayout`. I think the only place where it's used is in the envelope (if I didn't missed something).

write.go
```diff
@@ -177,8 +169,13 @@ func (w *Writer) writeField(field interface{}) error {
 		return w.writeLiteral(field)
 	case []interface{}:
 		return w.writeList(field)
+	case messageDateTime:
+		return w.writeDateTime(time.Time(field), messageDateTimeLayout)
+	case Date:
+		return w.writeDateTime(time.Time(field), DateLayout)
+	case DateTime:
 	case time.Time:
```
emersion: https://play.golang.org/p/yvuOXZI5sA

`case DateTime:` won't do anything here.

search.go
```diff
@@ -196,7 +205,7 @@ func (c *SearchCriteria) Format() (fields []interface{}) {
 		fields = append(fields, "BCC", c.Bcc)
 	}
 	if !c.Before.IsZero() {
-		fields = append(fields, "BEFORE", FormatDate(c.Before))
+		fields = append(fields, "BEFORE", c.Before.Format(DateLayout))
```
emersion: Either we choose to format dates here and we remove types `Date` and `DateTime`, either we choose to always use them. But I'd prefer not to mix both.

search_test.go
emersion: >you sure this is what you want the server to do?

What do you mean?

emersion:
```
# github.com/emersion/go-imap/client_test
client/cmd_selected_test.go:8: imported and not used: "time"
client/cmd_selected_test.go:113: undefined: imap.ParseDate
```

PR #75, opened 2016-12-01
Make ErrorLog a small interface

### LINKED ISSUE #74: Make ErrorLog an interface
Having `ErrorLog` as an interface would make replacement very easy.

ferhatelmas:
* which enables easier override for custom behavior
 * fixes #74


emersion:
LGTM, thanks!

PR #81, opened 2016-12-03
Add client Dial functions with net.Dialer parameters

horgh:
In these changes I add two client Dialer functions that let us pass in a net.Dialer. Primarily I wanted to make it possible to add timeouts.

This is similar to the crypto/tls package's DialWithDialer function: https://godoc.org/crypto/tls#DialWithDialer

It is likely possible to do something similar using the already present client.New function, but I figure this is a little bit of a nicer interface.

I also expose a way to retrieve the underlying imap.Conn/net.Conn to the client. Again this is because I want to be able to add timeouts (or rather, set a deadline).

Why? Well, I have a small program that runs from cron to check an IMAP mailbox. I found that sometimes the program hangs seemingly forever. Today I have 9 stuck running for example.

If I strace one of them I see this:

```
# strace -p 29433
strace: Process 29433 attached
futex(0x8537004, FUTEX_WAIT, 0, NULL
```

Looking at its open files:

```
# lsof -p 29433
COMMAND     PID USER   FD      TYPE   DEVICE SIZE/OFF     NODE NAME
imap-noti 29433 will  cwd       DIR      8,1    16384  2754342 /home/will
imap-noti 29433 will  rtd       DIR      8,1     4096        2 /
imap-noti 29433 will  txt       REG      8,1  7199880 28705989 /tmp/go-build337774699/summercat.com/imap-notify/_obj/exe/imap-notify (deleted)
imap-noti 29433 will  DEL       REG      8,1          23855716 /lib/i386-linux-gnu/libc-2.24.so
imap-noti 29433 will  DEL       REG      8,1          23855828 /lib/i386-linux-gnu/libpthread-2.24.so
imap-noti 29433 will  DEL       REG      8,1          23855499 /lib/i386-linux-gnu/ld-2.24.so
imap-noti 29433 will    0r     FIFO     0,10      0t0 46312492 pipe
imap-noti 29433 will    1u      REG      8,1        0 28705046 /tmp/tmpf8qQXZw (deleted)
imap-noti 29433 will    2u      REG      8,1        0 28705046 /tmp/tmpf8qQXZw (deleted)
imap-noti 29433 will    3u     IPv4 46312510      0t0      TCP leviathan.summercat.com:58476->pc-in-f108.1e100.net:imaps (CLOSE_WAIT)
imap-noti 29433 will    5u  a_inode     0,11        0     1036 [eventpoll]
```

I believe having a timeout will solve this.

If you are interested, in this commit I make use of the functions here to add timeouts/deadlines to my program:

https://github.com/horgh/imap-notify/commit/87cb58f786454b733a4740c41892914fa9c7c83c

Please let me know if you see anything I should improve or that you'd like me to change.

Thanks for your time!

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -1,4 +1,4 @@
-// An IMAP client.
+// Package client provides an IMAP client.
 package client
 
 import (
@@ -11,7 +11,7 @@ import (
 	"github.com/emersion/go-imap"
 )
 
-// An IMAP client.
+// Client represents an IMAP client.
 type Client struct {
 	conn  *imap.Conn
 	isTLS bool
@@ -343,7 +343,7 @@ func (c *Client) Upgrade(upgrader imap.ConnUpgrader) error {
 	return c.conn.Upgrade(upgrader)
 }
 
-// Get an imap.Writer for this client's connection.
+// Writer returns the imap.Writer for this client's connection.
 //
 // This function should not be called directly, it must only be used by
 // libraries implementing extensions of the IMAP protocol.
@@ -351,7 +351,7 @@ func (c *Client) Writer() *imap.Writer {
 	return c.conn.Writer
 }
 
-// Check if this client's connection has TLS enabled.
+// IsTLS checks if this client's connection has TLS enabled.
 func (c *Client) IsTLS() bool {
 	return c.isTLS
 }
@@ -392,7 +392,7 @@ func New(conn net.Conn) (c *Client, err error) {
 	return
 }
 
-// Connect to an IMAP server using an unencrypted connection.
+// Dial connects to an IMAP server using an unencrypted connection.
 func Dial(addr string) (c *Client, err error) {
 	conn, err := net.Dial("tcp", addr)
 	if err != nil {
@@ -417,7 +417,7 @@ func DialWithDialer(dialer *net.Dialer, address string) (c *Client, err error) {
 	return
 }
 
-// Connect to an IMAP server using an encrypted connection.
+// DialTLS connects to an IMAP server using an encrypted connection.
 func DialTLS(addr string, tlsConfig *tls.Config) (c *Client, err error) {
 	conn, err := tls.Dial("tcp", addr, tlsConfig)
 	if err != nil {
--- a/client/client.go
+++ b/client/client.go
@@ -3,10 +3,12 @@ package client
 
 import (
 	"crypto/tls"
+	"fmt"
 	"io"
 	"log"
 	"net"
 	"os"
+	"time"
 
 	"github.com/emersion/go-imap"
 )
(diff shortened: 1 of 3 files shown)


client/client.go
```diff
@@ -368,6 +368,11 @@ func (c *Client) SetDebug(w io.Writer) {
 	c.conn.SetDebug(w)
 }
 
+// Conn returns the client's imap.Conn.
+func (c *Client) Conn() *imap.Conn {
```
emersion: I'm not comfortable with a `Conn` method that returns an `*imap.Conn`.

Alternative ways of doing this:
* Use `client.New`, as you said in your comment
* Add `SetDeadline` function to `imap.Client`
* Add a `Timeout time.Duration` field to `imap.Client` (just like net/http)
* Add a `net.Conn` field to `imap.Client`
* Make this function return a `net.Conn`
* Another idea?

What do you think?

horgh: Thanks for looking at this and for the feedback!

Adding a Timeout field to imap.Client seems like it would be elegant. I am wondering though what it would really mean. It seems like to mean something similar to net/http it would mean setting a deadline for every command issued. Which might make this change a lot more complex.

I'm leaning towards adding a SetDeadline or changing this function to return net.Conn.

Having a function to access the net.Conn might be safest interface wise. We wouldn't be going too far down a path where the interface suggests that is the best way to handle timeouts (which adding a SetDeadline seems to imply). Having it return net.Conn might be the simplest bet (other than just using client.New!).

I'll do that tomorrow unless you have any further thoughts.

Thanks again.

emersion: I would also prefer the Timeout field. It's not that complicated as you could just add:
```go
if c.Timeout > 0 {
    c.conn.SetDeadline(time.Now().Add(c.Timeout))
}
```
in `executeCommand`.

horgh:
Hi again!

Thanks for the pointer about adding the `SetDeadline`. I added this in `execute` (you mentioned `executeCommand`, but I think this is the one you meant?).

After testing I found a couple cases that were not timing out as I expected even so. I mentioned them in the commit, but here is a little more information:

First, I found that if the server I connected to never issued a greeting, the client would never time out. (I was testing using nc as a server). I realized this was because by the time I set `Timeout` on the `Client` only after `New` returns, by which point we've already started trying to read.

To get around this, I added a `SetDeadline` in the two `DialWithDialer` functions. I think this is kind of hacky and maybe surprising but I am not sure a better solution without changing the function interfaces which I expect is something you would rather avoid. Even with this, we need to ensure `New` knows we've failed, so I close the `greeted` channel.

Second, in my test server, I did the following:

```
S: Send greeting
C: Send LOGIN
S: <nothing>
C: <times out>
C: Send LOGOUT
S: * BYE etc
S: Acknowledge LOGOUT
```

I found that the client would time out at `LOGIN` and `read` would end. Then LOGOUT's `execute` would get stuck waiting for a status response which would never come, and the client hung forever.

To resolve this I added a check for whether the `loggedOut` channel is closed inside the `select` in `execute`.

Sorry for complicating this!

Thanks again for your time.

emersion:
Thanks for finding all these corner cases!

I've added a commit to prevent `Logout` from failing if the server is closing its connection (since that's what we want).

Lets merge this :+1:

PR #84, opened 2016-12-07
Simplify showcased client code

manishrjain:
Don't need the done channel because messages channel is being closed by the call to client.Fetch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,18 +90,15 @@ func main() {
 
 	// List mailboxes
 	mailboxes := make(chan *imap.MailboxInfo)
-	done := make(chan error, 1)
 	go func () {
-		done <- c.List("", "*", mailboxes)
+		if err := c.List("", "*", mailboxes); err != nil {
+			log.Fatal(err)
+		}
 	}()
 
 	log.Println("Mailboxes:")
 	for m := range mailboxes {
 		log.Println("* " + m.Name)
-	}
-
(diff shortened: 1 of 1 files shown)


emersion:
Can you do the same thing for `mailboxes`?

manishrjain:
Done. Please have another look!

README.md
```diff
@@ -115,20 +112,18 @@ func main() {
 	seqset, _ := imap.NewSeqSet("")
 	seqset.AddRange(mbox.Messages - 3, mbox.Messages)
 
-	messages := make(chan *imap.Message)
-	done = make(chan error, 1)
+	messages := make(chan *imap.Message, 10)
```
emersion: I didn't realized when I first looked that you used a buffered channel here. Is there a particular reason?

The `for msg := range messages` loop should process messages as soon as they arrive.

manishrjain:
In an unbuffered channel, each put into channel would block until there's an exact get from the channel. It basically makes both put and get synchronous. So, instead of producer working independently, concurrently from the receiver, they block on each other -- which isn't a great design.

Honestly, I think there are very few uses of unbuffered channels -- and they're very prone to deadlocks.

README.md
manishrjain: Answered directly in the PR.

emersion:
Okay, this makes sense, thanks for explaining!

LGTM

PR #90, opened 2016-12-24
Use a mutex to protect access the client Caps map

horgh:
Hi again!

This is to resolve a panic I encountered:

```
panic: assignment to entry in nil map

goroutine 37 [running]:
panic(0x8253380, 0x18768800)
        /home/will/bin/go-src/src/runtime/panic.go:500 +0x331
github.com/emersion/go-imap/client.(*Client).gotStatusCaps(0x189f8fa0, 0x18788798, 0x14, 0x15)
        /home/will/code/go/src/github.com/emersion/go-imap/client/client.go:234 +0x12c
github.com/emersion/go-imap/client.(*Client).handleUnilateral(0x189f8fa0)
        /home/will/code/go/src/github.com/emersion/go-imap/client/client.go:314 +0xbc5
created by github.com/emersion/go-imap/client.New
        /home/will/code/go/src/github.com/emersion/go-imap/client/client.go:435 +0x3c2
```

From what I can gather, there are at least two goroutines that will access this map. `handleUnilateral()`, and when using functions such as `Login()`.

Thanks for your time!

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -28,11 +28,11 @@ type Client struct {
 	greeted   chan struct{}
 	loggedOut chan struct{}
 
-	// The Caps map may be accessed in different goroutines. Protect access.
+	// The cached server capabilities.
+	caps map[string]bool
+	// The caps map may be accessed in different goroutines. Protect access.
 	capsLocker sync.Mutex
 
-	// The server capabilities.
-	Caps map[string]bool
 	// The current connection state.
 	State imap.ConnState
 	// The selected mailbox, if there is one.
@@ -234,11 +234,10 @@ func (c *Client) handleContinuationReqs(continues chan<- bool) {
 func (c *Client) gotStatusCaps(args []interface{}) {
 	c.capsLocker.Lock()
 
-	c.Caps = make(map[string]bool)
-
+	c.caps = make(map[string]bool)
 	for _, cap := range args {
 		if cap, ok := cap.(string); ok {
-			c.Caps[cap] = true
+			c.caps[cap] = true
 		}
 	}
 
--- a/client/client.go
+++ b/client/client.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"net"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/emersion/go-imap"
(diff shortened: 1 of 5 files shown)


emersion:
Hi!

I've been more and more willing to unexport `Client.Caps` and replace it with a `Client.Supports` function. What do you think? Since the API is not stable yet, I think we can change that before it becomes a bigger problem. This `Client.Caps` field has already been a problem in the past because it could be nil. Now we need to add a locker, which is a bad idea if the field is exported.

If it's okay for you, we need to define the signature of the new `Client.Supports` function.
* Maybe `Supports(cap string) bool`, which fetches automatically capabilities if the server didn't sent them, ignores errors and returns true if cap is supported
* Maybe `Supports(cap string) (bool, error)`
* Maybe `Supports(cap string) (supported bool, ok bool)` which does not fetch automatically capabilities but sets ok to false if capabilities are unknown (I don't like this one)

Or maybe you have a better idea?

horgh:
You're right, it makes little sense to export Caps while requiring a mutex like this. I missed that!

Of the 3 `Support()` interfaces you suggest, I like the second. It will always be good to know if there is an error or we could not determine the capabilities for some reason.

Would you want to alter `Capability()` too? It looks similar to what you are suggesting, other than that it does not cache caps. I suppose the caching is the key difference and main use of Caps.

I think doing this makes sense. I don't have a better suggestion, although I am wondering about whether it is necessary for it to make the Capability request if it is not populated. I'm not too sure the implications. I expect you know better!

If you like I can take a shot at implementing the second, or if you'd prefer to do so yourself, please feel free!

emersion:
>Of the 3 Support() interfaces you suggest, I like the second.

Yeah, it's also my preferred one, let's do this.

>Would you want to alter `Capability()` too? [...] I am wondering about whether it is necessary for it to make the Capability request if it is not populated.

So the idea would be to have two functions:
* `Capability()` which always sends a CAPABILITY command
* `Supports()` which sends a CAPABILITY command only if necessary

`Capability()` could be required to force to refresh capabilities. What are you suggesting exactly? I'm not sure I understand well. If capabilities are not available, what should the `Supports()` and `Capability()` functions do?

>If you like I can take a shot at implementing the second, or if you'd prefer to do so yourself, please feel free!

I can work on that tomorrow if you want. As you prefer!

emersion:
I added `Client.Support`, let me know what you think!

(I also renamed `SupportsAuth` to `SupportAuth` and `SupportsStartTLS` to `SupportStartTLS` as it seems that stripping the s is more idiomatic in Go)

horgh:
Thanks for doing this!

I looked over the changes and I can't see any problems. Looking good to me!

emersion:
Okay, let's merge this then! Thanks for your help!

PR #99, opened 2017-01-13
Panic due to closing closed channel in client execute()

horgh:
Hello again!

I encountered this panic:

```
panic: close of closed channel

goroutine 36 [running]:
panic(0x82533a0, 0x18a129d0)
        /home/will/bin/go-src/src/runtime/panic.go:500 +0x331                                 github.com/emersion/go-imap/client.(*Client).execute.func2(0x187f66c0, 0x8361c20, 0x187fc120,
0x187fc140)
        /home/will/code/go/src/github.com/emersion/go-imap/client/client.go:161 +0x71
created by github.com/emersion/go-imap/client.(*Client).execute
        /home/will/code/go/src/github.com/emersion/go-imap/client/client.go:164 +0x3cc
```

I've solved this by wrapping the call to `close(hdlr)` in a `sync.Mutex`.

However, I wonder if we need to actually close it in two places. In the first spot (after `done <- res.HandleFrom(hdlr)`), do we actually need to close it? In that case `HandleFrom` returned, so I think closing the channel isn't necessary any longer as `HandleFrom` is the only spot I see using the channel.

Thanks for your help!

Oh I should mention in case you need more info on how this happened (I can't reproduce it), this happened in my [imap-notify](https://github.com/horgh/imap-notify) program, which uses the client to `Login()`, `Select()`, `Fetch()` and `Logout()`. Let me know if I can tell you anything more.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -150,6 +150,8 @@ func (c *Client) execute(cmdr imap.Commander, res imap.RespHandlerFrom) (status 
 
 	var hdlr imap.RespHandler
 	var done chan error
+	// We may close hdlr from two goroutines. To do so safely, use a mutex.
+	hdlrLocker := &sync.Mutex{}
 	if res != nil {
 		hdlr = make(imap.RespHandler)
 		done = make(chan error, 1)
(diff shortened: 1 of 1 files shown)


emersion:
Thanks a lot for your PR! Actually you're right, closing the channel from two places was a bad idea from the beginning, so I made one goroutine only read from the channel, and the other only write (see 0140ce12085c20039800a3f8dade1695e1e1afea). The function is now much simpler.

More work needs to be done concerning race conditions, I'll try to work on this.

PR #123, opened 2017-07-04
Release v1

### LINKED ISSUE #106: OK [UNSEEN] doesn't return the sequence number of the first unseen message
See https://github.com/emersion/go-imap/issues/105#issuecomment-280861897

>The `OK [UNSEEN]` response returns the sequence number of the first unseen message, but the `STATUS` command returns the number of unread messages.

### LINKED ISSUE #107: Deadlock when executing commands in a loop
@emersion 
Hello,
I want to count number of messages in folders(mailboxes).
I made infinity loop only for tests. 

It's look like my problem from IDLE, that last command is not finished and next is ignored causing to block.

It's code from example, I only added code inside for reading from channel. 
```go
    for {

        c, err := client.Dial("localhost:143")
        if err != nil {
            log.Fatal(err)
        }
        log.Println("Connected")

        if err := c.Login("user", "password"); err != nil {
            panic(err)
        }
        log.Println("Logged in")

        // List mailboxes
        mailboxes := make(chan *imap.MailboxInfo, 10)
        go func() {
        if err := c.List("", "*", mailboxes); err != nil {
                log.Fatal(err)
            }
        }()

        log.Println("Mailboxes:")
        for m := range mailboxes {
            log.Println("* " + m.Name)

            mbox, err := c.Select(m.Name, true)
            if err != nil {
                log.Println("Error", err)
                return
            }
            fmt.Println(m.Name, " + ", mbox.Messages)
            c.Close()

        }

        c.Logout()
    }
```

```
E336w OK [READ-ONLY] Examine completed (0.000 secs).
* OK [CLOSED] Previous mailbox closed.
* FLAGS (\Answered \Flagged \Deleted \Seen \Draft)
* OK [PERMANENTFLAGS ()] Read-only mailbox.
* 26 EXISTS
* 0 RECENT
* OK [UIDVALIDITY 1484216700] UIDs valid
* OK [UIDNEXT 27] Predicted next UID
* OK [NOMODSEQ] No permanent modsequences
36w OK [READ-ONLY] Examine completed (0.000 secs).
imap/client: 2017/02/23 11:00:15 response has not been handled: &{* OK NOMODSEQ [] No permanent modsequences}
lE336w EXAMINE Sent
36w EXAMINE Sent
```

### LINKED ISSUE #121: client: make Client.State and Client.Mailbox getters
`State` and `Mailbox` are struct fields for now. Converting them to getters has two advantages:

* It prevents the user from modifying them
* It allows the library to prevent race conditions - `State` is written by `handleUnilateral` and read by `read`

### LINKED ISSUE #124: client: rewrite tests
Client tests use two goroutines, which makes it impossible to use `t.Errorf` in one of them.

### LINKED ISSUE #136: Interaction between EXPUNGE and other responses
Hi, first, many thanks indeed for a terrific library! :smile:

I've got a question about how to handle server responses in correct order using the go-imap client.  I'm concerned because EXPUNGE responses are sent to a different golang channel than, for example, FETCH responses.  I've seen an example with gmail where after sending a MOVE command to the server, both an EXPUNGE and a FETCH response (flags updated on some other message by another client) happen to be returned.  Clearly, the responses arrive in a defined order, and local state must be updated in that order otherwise corruption/de-synchronisation may occur.  But, by sending these responses to c.Expunges and c.MessageUpdates, I don't then know how to interpret them in the order that arrived, unless I make all these channels non-buffered, and have a channel reader selecting across these and the other relevant channels to reconstruct the ordering.  Is that the intended implementation pattern?  If so, it's moderately painful and non-obvious.  Would it make more sense to have a single channel to handle all server responses in order?

emersion:
This is going to break a lot of stuff.

Milestone: https://github.com/emersion/go-imap/milestone/1

- [x] Use separate types instead of `string`?
- [ ] Race conditions #151
- [ ] Update all extensions
    - [ ] APPENDLIMIT
    - [x] COMPRESS: no update required
    - [ ] ENABLE
    - [ ] ID
    - [x] IDLE: https://github.com/emersion/go-imap-idle/pull/9
    - [x] MOVE
    - [ ] QUOTA
    - [x] SPECIAL-USE: no update required
    - [ ] UNSELECT
    - [ ] UIDPLUS
- [ ] Update all backends
  - [x] Memory
  - [ ] Multi
  - [ ] PGP
  - [x] Proxy

# imap

* `NewSeqSet` has been removed, use `new(SeqSet)` or `ParseSeqSet` instead
* `MimeSpecifier`, `BodyStructure.MimeType` and `BodyStructure.MimeSubType` have been renamed to `MIMESpecifier`, `BodyStructure.MIMEType` and `BodyStructure.MIMESubType`
* `BodyStructure.Md5` has been renamed to `BodyStructure.MD5`
* `RespHandler` and all response handling stuff has been removed and replaced by `responses.Handler`
* `*MsgAttr` constants have been replaced by `Fetch*` constants, functions that used to take a `string` now take a `FetchItem`
* `Mailbox*` constants have been replaced by the `StatusItem` type
* Command name constants have been removed
* `Status*` constants have been renamed to `StatusResp*` to prevent clash with the new `StatusItem` constants
* `Code*` constants are now of type `StatusRespCode`

# backend

* `Fetch` now takes a `[]imap.FetchItem` instead of a `[]string` for fetched items
* `Status` now takes a `[]imap.StatusItem` instead of a `[]string` for retrieved items

# client

* `State` and `Mailbox` are now getters
* `Infos`, `Warnings`, `Errors`, `Byes`, `MailboxUpdates`, `Expunges` and `MessageUpdates` channels have been replaced by one `Updates` channel
* `Fetch` now takes a `[]imap.FetchItem` instead of a `[]string` for fetched items
* `Store` now takes a `imap.StoreItem` instead of a `string` for the stored item
* `Status` now takes a `[]imap.StatusItem` instead of a `[]string` for retrieved items

# server

* `SaslServerFactory` has been renamed to `SASLServerFactory`

# utf7

* `Encoder` and `Decoder` globals have been replaced by `Encoding`

Fixes #121
Fixes #124
Fixes #107
Fixes #136
Fixes #106


emersion:
@hartshorne With #132, I think we're fine: only one goroutine is spawn to read responses from the server. All response handling is done in this same goroutine. I believe all calls to `ErrorLog.Print*` are made in this goroutine: if an error occurs when executing a command, it's directly returned.

PR #137, opened 2017-08-15
Replace various update channels with one unique channel

### LINKED ISSUE #136: Interaction between EXPUNGE and other responses

emersion:
- [x] Maybe create one `Update` struct per update type?

@jim-minter, what do you think?

Fixes #136

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -51,7 +51,8 @@ type Client struct {
 	// A channel to which unilateral updates from the server will be sent. An
 	// update can be one of: *imap.StatusResp, *imap.MailboxStatus, *imap.Message,
 	// *ExpungeUpdate. Note that blocking this channel blocks the whole client,
-	// so it's recommended to use a buffered channel.
+	// so it's recommended to use a separate goroutine and a buffered channel to
+	// prevent deadlocks.
 	Updates chan<- interface{}
 
 	// ErrorLog specifies an optional logger for errors accepting
--- a/client/client.go
+++ b/client/client.go
@@ -22,9 +22,24 @@ var errClosed = fmt.Errorf("imap: connection closed")
 // errUnregisterHandler is returned by a response handler to unregister itself.
 var errUnregisterHandler = fmt.Errorf("imap: unregister handler")
 
+// StatusUpdate is delivered when a status update is received.
+type StatusUpdate struct {
+	Status *imap.StatusResp
+}
+
+// MailboxUpdate is delivered when a mailbox status changes.
+type MailboxUpdate struct {
+	Mailbox *imap.MailboxStatus
+}
+
 // ExpungeUpdate is delivered when a message is deleted.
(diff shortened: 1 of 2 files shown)


client/client.go
```diff
@@ -291,22 +286,10 @@ func (c *Client) handleUnilateral() {
...
-						c.Warnings <- resp
-					}()
-				}
-			case imap.StatusBad:
-				if c.Errors != nil {
+			case imap.StatusOk, imap.StatusNo, imap.StatusBad:
+				if c.Updates != nil {
 					go func() {
-						c.Errors <- resp
+						c.Updates <- resp
```
jim-minter: I don't think it's legitimate to put this in a separate go-routine: if sending to the channel blocks and multiple goroutines exist trying to send different items, isn't the channel item ordering undefined again?
I think the library has to block here, and the user must be responsible for consuming from the channel in parallel with the function call (or somehow be very confident their channel buffer size is large enough to avoid deadlock).

jim-minter: (this comment applies throughout this PR)

emersion: You're right, fixed!

jim-minter:
Looks good - many thanks @emersion!

client/client.go
```diff
@@ -43,20 +48,11 @@ type Client struct {
...
-	// A channel where mailbox updates from the server will be sent.
-	MailboxUpdates chan *imap.MailboxStatus
-	// A channel where deleted message IDs will be sent.
-	Expunges chan uint32
-	// A channel where messages updates from the server will be sent.
-	MessageUpdates chan *imap.Message
+	// A channel to which unilateral updates from the server will be sent. An
+	// update can be one of: *imap.StatusResp, *imap.MailboxStatus, *imap.Message,
+	// *ExpungeUpdate. Note that blocking this channel blocks the whole client,
+	// so it's recommended to use a buffered channel.
```
emersion: Yeah, that's a good point, I changed the docs. In general, it's unsafe to rely on the server only sending a given number of responses and it's better to use separate goroutines.

PR #138, opened 2017-08-17
fix Updates type

jim-minter:
Updates must be of type `chan interface{}`, not `chan<- interface{}`, otherwise clients can't read from the channel :)


emersion:
I'm not sure about this. The intended usage is:

```go
ch := make(chan interface{}, 10)
c.Updates = ch
go receiveUpdates(ch)
```

This ensures that the library only sends to the channel.

jim-minter:
Ah, I see, thanks.

PR #139, opened 2017-09-03
quit sending when server is closed

ineiti:
I'm not sure this is the right thing to do - but when I use your excellent go-imap server in my tests, the `send()`-method keeps waiting, even after I call `Close()`. This patch closes the `send()`-method, so go-imap doesn't keep a go-routine after closing.

### DIFF
--- a/server/conn.go
+++ b/server/conn.go
@@ -198,20 +198,23 @@ func (c *conn) send() {
 
 	// Send responses
 	for {
-		// Get a response that needs to be sent
-		res := <-c.responses
+		select {
+		case res := <-c.responses:
+			// Get a response that needs to be sent
+			// Request to send the response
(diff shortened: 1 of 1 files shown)


emersion:
First, thanks for trying to fix this issue! :)

However, the `c.continues` channel handles continuation requests, so it's not the right channel to read from. I think another channel needs to be added - with the same purpose of the [`loggedOut`](https://github.com/emersion/go-imap/blob/master/client/client.go#L29) channel in the client. Do you want to try to add that or do you prefer that I look into it?

Also, make sure to use the v1 branch if possible. ;)

emersion:
Fixed in https://github.com/emersion/go-imap/commit/395a1a3bc89048049f09baa5204fbcbed414a972

PR #162, opened 2018-01-03
read: fix crlf reading

ferhatelmas:
Return after getting error on reading CR. Accordingly, updated fields reading.


emersion:
LGTM, thanks!

PR #164, opened 2018-01-05
client: fix couple of formatting issues in tests

ferhatelmas:
* formatted call but no format directive
* formatted call but no need
* intend to print value, but function was given


emersion:
Thanks!

PR #165, opened 2018-01-05
travis: make docker based, instead of vm

ferhatelmas:
brings startup speed


emersion:
I thought this was the default nowadays, but I can't find any info about it. I guess it's better to add it. Thanks!

ferhatelmas:
@emersion cool, it even increases coverage :wink:

PR #166, opened 2018-01-05
message: handle errors in bodystructure parsing

ferhatelmas:


ferhatelmas:
@emersion how about dividing that big function into smaller pieces and handling errors instead of intentional ignore ?

emersion:
Yeah, that would be way better!

ferhatelmas:
outdated

Issue #190, opened 2018-07-03
Problem with STORE command


Frabec:
Hi,
Thank you for your tool, I find it very useful. I've noticed recently that there is a problem with the STORE command. I use go-imap with go-imap-proxy to communicate through the proxy with an IMAP server (imap.gmail.com or my own dovecot server). When I communicate with the proxy using telnet  I log in I select a mailbox and then when I put the following command 
```
$ a STORE 2 -FLAGS (\Seen)
```
I get the following error : 
```
a NO imap: cannot format field: [\Seen]
```
And then the server freezes and I cannot logout I have to restart it

PR #191, opened 2018-07-04
write: Solved problem with Store command

Frabec:
Solves the problem of issue #190

### DIFF
--- a/write.go
+++ b/write.go
@@ -33,7 +33,7 @@ func formatNumber(num uint32) string {
 func FormatStringList(list []string) (fields []interface{}) {
 	fields = make([]interface{}, len(list))
 	for i, v := range list {
-		fields[i] = v
+		fields[i] = Atom(v)
 	}
 	return
 }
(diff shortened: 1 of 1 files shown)


emersion:
Thanks for you PR!

Would it be possible to instead change the string list to be an interface{} list? Otherwise we'd have to implement formatting for other list items too. You can use this function to do so: https://godoc.org/github.com/emersion/go-imap#FormatStringList

Frabec:
I had tried at first to do as you say i.e convert the string list to an interface{} list. The problem is when I do that my postfix server throws me this error : 
`NO Error in IMAP command STORE: Flags list contains non-atoms (0.001 + 0.000 secs)`

emersion:
Hmm. Can you try converting each `string` to an `imap.Atom`?

Frabec:
Cool! You are right, it works now. I've just modified your function FormatStringList()

emersion:
Hmm, I don't think this is right, because some other functions use `FormatStringList` and need escaping.

It seems that `client.Store` already has code to deal with this: https://github.com/emersion/go-imap/blob/v1/client/cmd_selected.go#L180

This makes me believe that this is actually a bug in go-imap-proxy. This function should convert the `[]string` into an `[]interface{}` containing strings: https://github.com/emersion/go-imap-proxy/blob/master/mailbox.go#L97

Frabec:
I've checked the type of `value` in function https://github.com/emersion/go-imap-proxy/blob/master/mailbox.go#L97 when using the proxy and it is a `[]string`. I've added a line to fix the bug in go-imap-proxy. I've made a PR. 

Thanks for your help

Issue #140, opened 2017-09-07
Error reading response: unexpected EOF


advancedlogic:
Hi,
this apparently happens on the client.go and the application gets stuck and must be restarted.
Sounds like there is a never ending for { } loop in some specific case.
```go
		res, err := imap.ReadResp(c.conn.Reader)
		if err == io.EOF || c.State == imap.LogoutState {
			return nil
		}
		if err != nil {
			c.ErrorLog.Println("error reading response:", err)
			if imap.IsParseError(err) {
				continue
			} else {
				return err
			}
		}
```

ihippik:
any news?)

emersion:
Sorry, I don't have time to investigate this issue right now. Please send a pull request fixing it!

emersion:
This has been fixed in https://github.com/emersion/go-imap/pull/198

PR #198, opened 2018-09-28
Client - FIXED: return err to chan, from c.read()

trueone:
According to issue [#140](https://github.com/emersion/go-imap/issues/140)

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -454,7 +454,9 @@ func (c *Client) handleGreetAndStartReading() error {
 
 	// Make sure to start reading after we have set up this handler, otherwise
 	// some messages will be lost.
-	go c.read(greeted)
+	go func() {
+		done <- c.read(greeted)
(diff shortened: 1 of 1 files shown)


client/client.go
```diff
@@ -454,7 +454,9 @@ func (c *Client) handleGreetAndStartReading() error {
 
 	// Make sure to start reading after we have set up this handler, otherwise
 	// some messages will be lost.
-	go c.read(greeted)
+	go func() {
+		done <- c.read(greeted)
```
emersion: We probably need to create the `done` channel with a buffer of 2, in case both this goroutine and the other one fails. Otherwise we might end up in a goroutine leak.

trueone:
Hi, thanks for ur reply! Changed channel buffer

emersion:
Thanks!

PR #200, opened 2018-10-15
backendutil: read body section parts left-to-right

crawshaw:
This matches the description and example in RFC 3501 Section 6.4.5.


emersion:
Thanks for your contribution!

PR #202, opened 2018-10-15
imap: quote parameter values

crawshaw:
An email from outlook.com contained an embedded part with
Content-ID and filename starting with a long sequence of
digits:

	279351849188E3D84D6DCC5B@namprd05.prod.outlook.com

When sent unquoted, the Apple Mail IMAP client attempts to
parse this initially as an integer. When it overflows a
uint32 parsing is stopped, so the message contents are never
displayed.

Quoting all parameter values is safe and avoids this.
Indeed the IMAP RFC includes examples of far more aggressive
quoting of almost all strings from a MIME message.
That may be worth considering in the future

### DIFF
--- a/message.go
+++ b/message.go
@@ -88,7 +88,7 @@ func ParseParamList(fields []interface{}) (map[string]string, error) {
 func FormatParamList(params map[string]string) []interface{} {
 	var fields []interface{}
 	for key, value := range params {
-		fields = append(fields, key, value)
+		fields = append(fields, key, Quoted(value))
 	}
 	return fields
 }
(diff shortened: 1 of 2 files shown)


emersion:
Hum, that's pretty weird. We have logic to [quote automatically what needs to be quoted](https://github.com/emersion/go-imap/blob/v1/write.go#L90). IIRC in some cases quoting when it's not necessary breaks some clients (for instance when sending a quoted `INBOX` literal for mailbox names).

crawshaw:
That logic does not catch this case. It is valid by the RFC for an atom to be made up of digits. There are however clients that trip over it (Apple Mail), and expect atoms that start with digits to be quoted.

emersion:
Do you think it would be better to always quote atoms that begin with a digit, instead of making this a special case?

emersion:
Let's just merge this for now

emersion:
Thanks!

Issue #186, opened 2018-05-28
How To Search


fipso:
Hello,
how can I use the imap SEARCH command ?

emersion:
Read the RFC, grep for SEARCH.

fipso:
Hello,
sry for the stupid question.
How can i use this libary to make the Request with the command and get the response ?

emersion:
There's an example on the GoDoc page.

fipso:
The Doc just explanes how to create a search certeria:
https://godoc.org/github.com/emersion/go-imap#SearchCriteria

emersion:
Here: https://godoc.org/github.com/emersion/go-imap/client#Client.Search

Sorry, I'm mistaken, there's no example (yet). You can read the tests to see some usage examples though.

fipso:
Thanks a lot ;D

Issue #172, opened 2018-02-09
how to fetch UNreaded (UNSEEN) emails ?


happyalexkg:
Can you please tell me how to get only UNREADed (UNSEEN) emails

Tried some methiod .. but did not get any successful result :(

emersion:
You'll first need to `Search` messages (with `SeenFlag` in `SearchCriteria.WithoutFlags`) and then you'll be able to `Fetch` them.

PR #204, opened 2018-10-17
Add an example for Client.Search() usage

pabluk:
I'm new in golang and some days ago I spent some hours trying to figure out how to use [Client.Search](https://godoc.org/github.com/emersion/go-imap/client#Client.Search) and finally I understand it but I also found that I wasn't the only one looking for that, see https://github.com/emersion/go-imap/issues/186#issuecomment-392602873 and https://github.com/emersion/go-imap/issues/172. So I decided to write a basic example of searching for unseen messages.

I try to follow the same style of other examples but please don't hesitate to tell if I need to change something and feel free to accept or reject this PR.

### DIFF
--- a/client/example_test.go
+++ b/client/example_test.go
@@ -218,3 +218,45 @@ func ExampleClient_Store() {
 
 	log.Println("Message has been marked as seen")
 }
+
+func ExampleClient_Search() {
+	// Let's assume c is a client
+	var c *client.Client
+
+	// Select INBOX
+	_, err := c.Select("INBOX", false)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Set search criteria
+	criteria := imap.NewSearchCriteria()
+	criteria.WithoutFlags = []string{imap.SeenFlag}
+	ids, err := c.Search(criteria)
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.Println("IDs found:", ids)
+
+	if len(ids) > 0 {
+		seqset := new(imap.SeqSet)
+		seqset.AddNum(ids...)
+
+		messages := make(chan *imap.Message, 10)
+		done := make(chan error, 1)
+		go func() {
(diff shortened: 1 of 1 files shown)


client/example_test.go
```diff
@@ -218,3 +218,46 @@ func ExampleClient_Store() {
...
+	criteria := imap.NewSearchCriteria()
+	criteria.WithoutFlags = []string{imap.SeenFlag}
+	ids, err := c.Search(criteria)
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.Println("IDs found:", ids)
+
+	if len(ids) > 0 {
+
```
emersion: Style: extraneous line here

client/example_test.go
```diff
@@ -218,3 +218,46 @@ func ExampleClient_Store() {
...
+		seqset := new(imap.SeqSet)
+		seqset.AddNum(ids...)
+
+		messages := make(chan *imap.Message, 10)
+		done := make(chan error, 1)
+		go func() {
+			done <- c.Fetch(seqset, []imap.FetchItem{imap.FetchEnvelope}, messages)
+		}()
+
+		log.Println("Last messages:")
```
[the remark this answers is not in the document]
pabluk: good catch! yes I'll change that.

client/example_test.go
pabluk: oh yes I'll change that. Weird thing gofmt didn't complain about that.

pabluk:
I've applied all the requested changes, I hope it's ok now.

Thanks for your time and feedback!

emersion:
Thanks!

zychappy:
it is a nice example which can println msg subject,but
when I try to get msg body,the result always nil
```go
var sect imap.BodySectionName
	for msg := range messages {
		fmt.Println("* " + msg.Envelope.Subject)
		r := msg.GetBody(&sect)
		if r == nil {
			panic("no body")
		}
...
```

PR #220, opened 2019-03-01
Accept LF as command separator

foxcpp:
Handled by most wide-spread server implementations. Makes debugging easier because you can use utilities like netcat to directly talk to server.

### DIFF
--- a/read.go
+++ b/read.go
@@ -158,6 +158,9 @@ func (r *Reader) ReadCrlf() (err error) {
 	if char, _, err = r.ReadRune(); err != nil {
 		return
 	}
+	if char == lf {
+		return
+	}
 	if char != cr {
 		err = newParseError("line doesn't end with a CR")
 		return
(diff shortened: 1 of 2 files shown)


foxcpp:
Woops, seems to break after
```
A LOGIN foxcpp hi
B SELECT INBOX
C FETCH 1:1 (FULL)
```

emersion:
Note that `netcat` has a `-C, --crlf` option.

foxcpp:
> Note that `netcat` has a `-C, --crlf` option.

Oh, really? Didn't known. I guess I'm trying to solve the problem on different end.
It wouldn't hurt much to merge it anyway or you have some concerns about test coverage, for example?

emersion:
>It wouldn't hurt much to merge it anyway or you have some concerns about test coverage, for example?

Oh yeah, now that it's fixed, this LGTM.

Thanks!

PR #223, opened 2019-03-03
Fix client data race in `client.execute()` method

Neopallium:
Need to always wait for the `cmd.WriteTo` to finish before returning
from from `client.execute()`, otherwise another goroutine could be
started before the current on finished.

This data race causes random `short write` errors.


emersion:
Thanks a lot!

Issue #151, opened 2017-11-10
Remaining race conditions


emersion:
- [ ] client `imap.(*Conn).SetDebug()` (do we care?)
- [ ] client `imap.(*Conn).Upgrade()`
- [ ] client `client.(*Client).Authenticate()`
- [ ] server continuation requests
- [ ] server conn upgrade

Neopallium:
I also found a race condition in client `client.(*Client).execute()`.  See PR #223 for a fix.

Neopallium:
Also fixed the `server continuation request` data race in PR #223.

The lock in `server.(*conn)` really isn't needed anymore, as long as only one goroutine can write to the connection.  Also don't need a separate goroutine for writing continuations.

Neopallium:
The `Upgrade()` data races need a better way to interrupt and block read/writes to the connection.

emersion:
Yeah, the upgrade mechanism is a little involved. I wonder how we could make it better.

Neopallium:
@emersion I have some code working to fix the upgrade data races.  Just need to clean it up before sending a PR.

Should be able to fix the `SetDebug()` data race too.  The client `SetDebug()` will need to send a `NOOP` command to unblock the reader goroutine.

emersion:
Nice!

`SetDebug` is less important, because it shouldn't happen in production. Would still be nice to fix of course. Not sure what you mean by sending `NOOP`, I think I'm missing some context.

Neopallium:
@emersion All remaining data races fixed in PR #223 .

To fix the `SetDebug()` data race we need to unblock the reader goroutine.  The only way I could think of to do that is to cause the server to send a response.  I used the upgrade logic and a `NOOP` command to fix the data race in `SetDebug()`.  Also added a `LockedWriter` to protect the debug `io.Writer` from concurrent writes.

For the `Authenicate` data race I added a `replies` channel for the auth response handler to pass the replies to the `client.execute()` method.  So only one goroutine is doing writes to the socket.

PR #225, opened 2019-03-06
Fixes client & server data races.

Neopallium:
Fixes the data races in issue #151 .

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"net"
 	"os"
+	"syscall"
 	"sync"
 	"time"
 
@@ -156,6 +157,13 @@ func (c *Client) read(greeted <-chan struct{}) error {
 		if err == io.EOF || c.State() == imap.LogoutState {
 			return nil
 		} else if err != nil {
+			if opErr, ok := err.(*net.OpError); ok {
+				if syscallErr, ok := opErr.Err.(*os.SyscallError); ok {
+					if syscallErr.Err == syscall.ECONNRESET {
+						return nil
+					}
+				}
+			}
 			c.ErrorLog.Println("error reading response:", err)
 			if imap.IsParseError(err) {
 				continue
--- a/client/client.go
+++ b/client/client.go
@@ -62,6 +62,7 @@ type Client struct {
 	conn  *imap.Conn
 	isTLS bool
 
+	replies   chan string
 	loggedOut chan struct{}
 	upgrading bool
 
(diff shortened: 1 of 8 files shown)


client/client.go
```diff
@@ -156,6 +157,13 @@ func (c *Client) read(greeted <-chan struct{}) error {
 		if err == io.EOF || c.State() == imap.LogoutState {
 			return nil
 		} else if err != nil {
+			if opErr, ok := err.(*net.OpError); ok {
+				if syscallErr, ok := opErr.Err.(*os.SyscallError); ok {
+					if syscallErr.Err == syscall.ECONNRESET {
```
emersion: How did you end up hitting this error? (Just wondering whether this is the right fix)

server/conn.go
```diff
@@ -293,6 +294,9 @@ func (c *conn) serve(conn Conn) error {
 					Type: imap.StatusRespBad,
 					Info: err.Error(),
 				}
+			} else if strings.Contains(err.Error(), "use of closed network connection") {
```
emersion: Hmm, not of fan of this...

client/client.go
Neopallium: It sometimes happens when running the client tests.

server/conn.go
Neopallium: I was getting a lot of those errors when running the server tests.

I think it only happens because the tests close the socket directly.

responses/authenticate.go
```diff
@@ -7,22 +7,21 @@ import (
 	"github.com/emersion/go-sasl"
 )
 
+type AuthReplyFunc func(reply string)
```
emersion: Can this return an error? (Even if we don't use it right now)

client/client.go
```diff
@@ -62,6 +62,7 @@ type Client struct {
 	conn  *imap.Conn
 	isTLS bool
 
+	replies   chan string
```
emersion: Can this take a `[]byte` instead of a `string`? This would make it more obvious that it directly writes to the socket,

emersion:
Overall this looks pretty good, thanks for working on this!

A clean standalone commit history also helps a lot reviewing, +1 :)

responses/authenticate.go
Neopallium: LOL.  The first version of that had returned `error`.  I will put it back.

client/client.go
Neopallium: ok

server/conn.go
Neopallium: I dropped this commit.

Neopallium:
@emersion I applied the changes you requested and dropped the `use of closed network connection` commit.

Neopallium:
I fixed the commit messages of the last two commits.  I need to be more careful passing the commit message from the cli.

emersion:
Thanks!

PR #231, opened 2019-03-31
 Always close ch chan *imap.Message upon fetch

### LINKED ISSUE #230: defer closing channel in fetch should be earlier
Hey guys,

As I'm quite new to golang, this took me a while to figure out. The problem is in package client, cmd_selected.go

Basically, I'm using UidFetch, passing in a channel to accept `*imap.Message`, then transforming them to my own model as I need to, and in my app, returning `chan *MyOwnMessageModel, chan error`

The problem is, I'd like to be able to read *multiple* errors from the error channel as there could be errors when parsing messages, not just client-related. So basically, when an ErrNoMailboxSelected error is returned, my channel passed to UidFetch is not being closed. I'm relying on the fact it'll be closed once the fetch is completed (either with an error or successfully), where I could also close my `out` channel.

I could for example introduce two error channels, read via UidFetch in goroutine, write the UidFetch error in a separate channel, then read from it in the goroutine that transforms the data read and then re-write the error in another channel, check what it is and close the UidFetch channel myself, but that seems like too much complexity given that closing this channel could be moved 3 lines up. Please note I'm returning the same channel created in my `Reader`'s UidFetch to which I'm writing the error returned by IMAP Client's UidFetch.

On a second thought, I guess I could make it simpler by getting the error returned by go-imap/client.Client.UidFetch() and assert its type to be `ErrNoMailboxSelected`, then close the `chan *imap.Message` channel myself, but that doesn't seem like a very clean usage and it was somewhat counter-intuitive to me, in this case, leaving me to wonder what's going on and where exactly my bug could be.

Is there any particular reason why deferred closing is below the `if c.State() != imap.SelectedState {` if statement, and is it possible to move it up and always close the `ch` channel?

```
func (c *Client) fetch(uid bool, seqset *imap.SeqSet, items []imap.FetchItem, ch chan *imap.Message) error {
	if c.State() != imap.SelectedState {
		return ErrNoMailboxSelected
	}

	defer close(ch)

	var cmd imap.Commander
	cmd = &commands.Fetch{
		SeqSet: seqset,
		Items:  items,
	}
	if uid {
		cmd = &commands.Uid{Cmd: cmd}
	}

	res := &responses.Fetch{Messages: ch}

	status, err := c.execute(cmd, res)
	if err != nil {
		return err
	}
	return status.Err()
}
```
Here's my code, to give you an idea what I'm trying to accomplish
```
func (r *Reader) UidFetch(seqSet *imap.SeqSet) (chan *imap.Message, chan error) {
	messages := make(chan *imap.Message, 5)
	errChan := make(chan error, 1)
	go func() {
		section := &imap.BodySectionName{}
		errChan <- r.IMAP.UidFetch(seqSet, []imap.FetchItem{section.FetchItem(), imap.FetchEnvelope, imap.FetchUid}, messages)
	}()

	return messages, errChan
}

func (r *Reader) ReadMessages(seqSet *imap.SeqSet) (chan *Message, chan error) {
	in, errChan := r.UidFetch(seqSet)

	out := make(chan *Message, 5)

	go func() {
		for {
			m, more := <-in
			if !more {
				close(out)
				return
			}

			message, err := convertImapMessageToMessage(m)

			if err != nil {
				errChan <- err
			} else {
				out <- message
			}
		}
	}()

	return out, errChan
}
```

Reading is done like so

```
read:
				for {
					select {
					case err := <-errorChan:
						if err != nil {
							log.Println(fmt.Sprintf("Error reading messages: %s", err.Error()))
						}
					case m, more := <-messages:
						if !more {
							break read
						}

						r.Write(m) // for the record, this writes them to a separate app, this service is just a supplement that's just way more stable written in Go than in PHP
					}
				}
```

PS Just figured that maybe I should probably close the `errChan` in addition to `out` when `in` is drained? Guess that won't fix the issue at hand though.

PPS I've already read through #131 and tried v1 beta, but mentioned code appears to be the same.

wucdbm:
Fixes #230

### DIFF
--- a/client/cmd_selected.go
+++ b/client/cmd_selected.go
@@ -134,11 +134,11 @@ func (c *Client) UidSearch(criteria *imap.SearchCriteria) (uids []uint32, err er
 }
 
 func (c *Client) fetch(uid bool, seqset *imap.SeqSet, items []imap.FetchItem, ch chan *imap.Message) error {
+	defer close(ch)
+
 	if c.State() != imap.SelectedState {
 		return ErrNoMailboxSelected
 	}
-
-	defer close(ch)
 
 	var cmd imap.Commander = &commands.Fetch{
 		SeqSet: seqset,
--- a/client/cmd_selected_test.go
+++ b/client/cmd_selected_test.go
@@ -240,6 +240,34 @@ func TestClient_Fetch(t *testing.T) {
 	}
 }
 
+func TestClient_Fetch_ClosedState(t *testing.T) {
+	c, s := newTestClient(t)
+	defer s.Close()
+
+	setClientState(c, imap.AuthenticatedState, nil)
+
+	seqset, _ := imap.ParseSeqSet("2:3")
+	fields := []imap.FetchItem{imap.FetchUid, imap.FetchItem("BODY[]")}
+
+	done := make(chan error, 1)
+	messages := make(chan *imap.Message, 2)
+	go func() {
+		done <- c.Fetch(seqset, fields, messages)
+	}()
+
+	_, more := <- messages
+
+	if more {
+		t.Fatalf("Messages channel has more messages, but it must be closed with no messages sent")
+	}
+
+	err := <- done
+
+	if err != ErrNoMailboxSelected {
+		t.Fatalf("Expected error to be IMAP Client ErrNoMailboxSelected")
+	}
+}
+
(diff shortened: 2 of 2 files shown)


emersion:
(I've rebased the changes)

emersion:
Thanks!

PR #234, opened 2019-04-08
Fix: send on closed channel

thanhps42:


thanhps42:
Sorry, I don't know why, but I got panic there :(

thanhps42:
panic: send on closed channel

goroutine 132 [running]:
github.com/emersion/go-imap/responses.(*Fetch).Handle(0xc0006dc000, 0x1cf5580, 0xc0001f6c90, 0x0, 0x2)
        D:/Go/src/github.com/emersion/go-imap/responses/fetch.go:34 +0x2f3
github.com/emersion/go-imap/client.(*Client).execute.func1(0x1cf5580, 0xc0001f6c90, 0xc000760160, 0xc0ffffffff)
        D:/Go/src/github.com/emersion/go-imap/client/client.go:230 +0x111
github.com/emersion/go-imap/responses.HandlerFunc.Handle(0xc00016e100, 0x1cf5580, 0xc0001f6c90, 0xc000760180, 0x0)
        D:/Go/src/github.com/emersion/go-imap/responses/responses.go:27 +0x4f
github.com/emersion/go-imap/client.(*Client).handle(0xc000760120, 0x1cf5580, 0xc0001f6c90, 0x0, 0x0)
        D:/Go/src/github.com/emersion/go-imap/client/client.go:114 +0x137
github.com/emersion/go-imap/client.(*Client).readOnce(0xc000760120, 0x1b01a01, 0x0, 0x0)
        D:/Go/src/github.com/emersion/go-imap/client/client.go:164 +0x2e1
github.com/emersion/go-imap/client.(*Client).reader(0xc000760120)
        D:/Go/src/github.com/emersion/go-imap/client/client.go:131 +0x86
created by github.com/emersion/go-imap/client.(*Client).handleGreetAndStartReading
        D:/Go/src/github.com/emersion/go-imap/client/client.go:498 +0x2f2

emersion:
What does your code look like?

thanhps42:
```
func (this *Hotmail) GetEmailsSubject(email, password string) ([]string, error) {
	var err error
	this.imapClient, err = client.DialTLS("outlook.office365.com:993", nil)
	if err != nil {
		return err
	}
	defer this.imapClient.Close()

	if err = this.imapClient.Login(email, password); err != nil {
		return err
	}
	defer this.imapClient.Logout()

	mbox, err := this.imapClient.Select("INBOX", true)
	if err != nil {
		return nil, err
	}

	from := uint32(1)
	to := mbox.Messages
	if mbox.Messages > 10 {
		from = mbox.Messages - 10
	}

	seqset := &imap.SeqSet{}
	seqset.AddRange(from, to)

	messages := make(chan *imap.Message, 20)
	if err = this.imapClient.Fetch(seqset, []imap.FetchItem{imap.FetchEnvelope}, messages); err != nil {
		return nil, err
	}

	var subjects []string
	for msg := range messages {
		if msg.Envelope == nil {
			continue
		}
		subjects = append(subjects, msg.Envelope.Subject)
	}

	return subjects, nil
}
```

emersion:
Not sure what happens here. `messages` should only be closed after the command has completed.

Neopallium:
maybe a data race?  try running with the `-race` flag.
`go run -race client.go`

Neopallium:
@emersion The channel is being closed in `Client.fetch()` instead of the response handler where results are sent from.  So there might be a data race.  It would be better to only close the channel from the same goroutine that is sending to it.

@thanhps42 It would help if you can catch the data race using the `-race` flag.

PR #235, opened 2019-04-14
Pass basic connection information to backend

### LINKED ISSUE #232: Expose connection state to the backend
Adding something similar to ConnectionState from go-smtp (https://github.com/emersion/go-smtp/issues/15) would be useful. For example, then [maddy](https://github.com/emersion/maddy) will be able to log IP addresses of clients on failed authentication, which is required for fail2ban to work.

_I guess I can implement it, just creating issue so I will not forget about it (I already did it like, 2 times)_

foxcpp:
Similar to go-smtp's ConnectionState. For https://github.com/emersion/maddy/issues/29. Breaking change.

Closes #232.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -2,6 +2,7 @@ package imap
 
 import (
 	"bufio"
+	"crypto/tls"
 	"io"
 	"net"
 	"sync"
@@ -53,8 +54,8 @@ const (
 type ConnUpgrader func(conn net.Conn) (net.Conn, error)
 
 type Waiter struct {
-	start sync.WaitGroup
-	end sync.WaitGroup
+	start    sync.WaitGroup
+	end      sync.WaitGroup
 	finished bool
 }
 
@@ -62,7 +63,7 @@ func NewWaiter() *Waiter {
 	w := &Waiter{finished: false}
 	w.start.Add(1)
 	w.end.Add(1)
-	return  w
+	return w
 }
 
 func (w *Waiter) Wait() {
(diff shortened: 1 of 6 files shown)


conn.go
```diff
@@ -135,6 +136,15 @@ func newMultiFlusher(flushers ...flusher) flusher {
 	return &multiFlusher{flushers}
 }
 
+// Underlying connection state information.
+type ConnInfo struct {
+	RemoteAddr net.Addr
+	LocalAddr  net.Addr
+
+	// Not populated if connection is not using TLS.
+	TLS tls.ConnectionState
```
emersion: Should be a pointer?

conn.go
```diff
@@ -135,6 +136,15 @@ func newMultiFlusher(flushers ...flusher) flusher {
 	return &multiFlusher{flushers}
 }
 
+// Underlying connection state information.
+type ConnInfo struct {
```
emersion: Should be named `ConnectionState`

foxcpp: There is already ConnState so I decided to pick a different name to prevent possible confusion. Should I still rename it to ConnectionState?

emersion: Gah, okay, makes sense

conn.go
foxcpp: Changed to pointer in ed492a9.

foxcpp:
CI broke :(

emersion:
Yeah. On the master branch I switched to sr.ht, but I guess this branch doesn't have the `.build.yml` file. Can you rebase?

foxcpp:
Rebased.

emersion:
Thanks!

PR #240, opened 2019-05-03
Handle HEADER.FIELDS in backendutil.FetchBodySection

foxcpp:


backend/backendutil/body.go
```diff
@@ -40,6 +40,20 @@ func FetchBodySection(e *message.Entity, section *imap.BodySectionName) (imap.Li
...
 
+	if section.Fields != nil {
+		if section.NotFields {
+			for _, fieldName := range section.Fields {
+				e.Header.Del(fieldName)
+			}
+		} else {
+			newHdr := make(message.Header)
+			for _, fieldName := range section.Fields {
+				newHdr.Set(fieldName, e.Header.Get(fieldName))
```
emersion: Should be `Add` in case there are duplicate header keys (`e.Header.Get` won't work in case of a duplicate). Since headers are inserted at the beginning, `section.Fields` should be iterated in reverse order.

emersion: TBH not sure what happens in case of duplicate header keys, should read again the RFC

foxcpp: https://tools.ietf.org/html/rfc5322#section-4.5
>Syntactically, the primary difference in the obsolete field syntax is
   that it allows multiple occurrences of any of the fields and they may
   occur in any order.

https://tools.ietf.org/html/rfc5322#section-4.5.3
>When multiple occurrences of destination address fields occur in a
   message, they SHOULD be treated as if the address list in the first
   occurrence of the field is combined with the address lists of the
   subsequent occurrences by adding a comma and concatenating.

nothing said about other fields tho..

foxcpp: Reversed iteration order in 17282a6. Not sure what to do with multiple values tho.

foxcpp: Something like that would work but looks hack-ish to me, what do you think?
```go
fields := e.Header[textproto.CanonialMIMEHeader(section.Fields[i])]
if fields == nil {
	continue
}
                                                                    
for j := len(fields) - 1; j >= 0; j-- {
	newHdr.Add(section.Fields[i], fields[j])
}
```

emersion: Hmm. Deleting keys that are not in `section.Fields` would probably be better, what do you think?

foxcpp: Yeah, that would look better.

emersion:
Thanks!

PR #241, opened 2019-05-04
backendutil: Merge Match* functions

### LINKED ISSUE #222: Unexpected result of MatchFlags for SearchCriteria with Not
Here is minimal test case:
```
import (
	"fmt"

	"github.com/emersion/go-imap"
	"github.com/emersion/go-imap/backend/backendutil"
)

func main() {
	crit := imap.SearchCriteria{
		Not: []*imap.SearchCriteria{
			&imap.SearchCriteria{
				Body: []string{"hi!"},
			},
		},
	}

	fmt.Println(backendutil.MatchFlags([]string{"FlagA", "FlagB"}, &crit))
}
```

Since there is no With(out)Flags I expect MatchFlags to return true but it returns false instead:
```
⟩ go run test.go
false
```

foxcpp:
Backward-compatibility-breaking change.
Closes #222.

### DIFF
--- a/backend/backendutil/search_test.go
+++ b/backend/backendutil/search_test.go
@@ -10,8 +10,14 @@ import (
 	"github.com/emersion/go-message"
 )
 
+var testInternalDate = time.Unix(1483997966, 0)
+
 var matchTests = []struct {
 	criteria *imap.SearchCriteria
+	seqNum   uint32
+	uid      uint32
+	date     time.Time
+	flags    []string
 	res      bool
 }{
 	{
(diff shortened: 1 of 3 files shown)


backend/backendutil/search_test.go
```diff
@@ -10,8 +10,14 @@ import (
...
 )
 
+var testInternalDate = time.Unix(1483997966, 0)
+
 var matchTests = []struct {
 	criteria *imap.SearchCriteria
+	seqNum   uint32
+	uid      uint32
+	date     time.Time
+	flags    []string
```
emersion: Style: alignment issue here

foxcpp: Ugh, gofmt seems to be fine with it. How it should be formatted?

emersion: Oh, if gofmt is fine with that, let's keep it that way

emersion: Argh, I see: for some reason GitHub displays `fl` as a ligature in one character. Thus `[]string` is misaligned. That's pretty weird.

emersion:
Thanks!

Issue #243, opened 2019-05-11
Run ImapTest


emersion:
https://github.com/dovecot/imaptest

foxcpp:
Tried to run imaptest stress testing against maddy. 
Using go-imap 1.0.0.beta.4 and go-imap-sql ([dev](https://github.com/foxcpp/go-imap-sql/tree/dev) branch).

Command:
```
imaptest seed=666 port=9999 copybox=INBOX2 user=testuser pass=foo mbox=~/Downloads/firefox/dovecot-crlf
```
Mentioned mbox file is here: http://www.dovecot.org/tmp/dovecot-crlf.

Outputer after running is two seconds: [imaptest_stress.log](https://github.com/emersion/go-imap/files/3179754/imaptest_stress.log) (look for "Error:" lines)

Looks like there is some issue that makes responses get intermixed with literals in responses.
Adding `cilents=1` leaves only
```
Error: testuser[300]: DELETE failed: 300.6 NO Flags must be a list
```

foxcpp:
Need to check against RFC if this is a valid syntax.
```
*** Test close command 1/4 (line 3)
 - failed: Expected tagged reply 'ok', got 'NO Flags must be a list'
 - Command (tag 4.13): store 1,3 +flags \deleted
```

foxcpp:
Relevant ABNF from RFC 3501.
```

flag-list       = "(" [flag *(SP flag)] ")"
flag            = "\Answered" / "\Flagged" / "\Deleted" /
                  "\Seen" / "\Draft" / flag-keyword / flag-extension
                    ; Does not include "\Recent"
flag-extension  = "\" atom
flag-keyword    = atom
store           = "STORE" SP sequence-set SP store-att-flags
store-att-flags = (["+" / "-"] "FLAGS" [".SILENT"]) SP
                  (flag-list / (flag *(SP flag)))
```

foxcpp:
Adding to the v2 milestone. There is number of issues detected by imaptest that could be resolved using only breaking changes (see #323).

PR #248, opened 2019-05-21
Handle "STORE FLAGS seqset flag1 flag2" syntax

foxcpp:
See https://github.com/emersion/go-imap/issues/243#issuecomment-492411070.

ABNF:
```
store-att-flags = (["+" / "-"] "FLAGS" [".SILENT"]) SP
                  (flag-list / (flag *(SP flag)))
```


emersion:
Thanks!

PR #249, opened 2019-05-21
Use less restrictve Dialer interface in DialWithDialer

### LINKED ISSUE #201: How would you use Socks5 dialer?
Is there a way implemented to connect via socks5?

foxcpp:
Closes #201.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -12,8 +12,8 @@ import (
 	"log"
 	"net"
 	"os"
-	"syscall"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/emersion/go-imap"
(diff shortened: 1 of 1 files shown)


foxcpp:
Woops.

foxcpp:
Okay, I fixed it.

client/client.go
```diff
@@ -592,11 +592,16 @@ func Dial(addr string) (c *Client, err error) {
 	return
 }
 
+type Dialer interface {
+	// Dial connects to the given address via the proxy.
+	Dial(network, addr string) (c net.Conn, err error)
```
emersion: We can remove the name of the return values here, they don't provide any documentation benefits.

foxcpp: Done!

emersion:
Thanks!

PR #250, opened 2019-05-21
Check literal length

### LINKED ISSUE #247: Check literal length
Right now when writing a literal we blindly `io.Copy` the `io.Reader`. We should make sure the `io.Reader` contains exactly `Len` bytes.

foxcpp:
Closes #247.

### DIFF
--- a/write.go
+++ b/write.go
@@ -151,8 +151,18 @@ func (w *Writer) writeLiteral(l Literal) error {
 		}
 	}
 
-	_, err := io.Copy(w, l)
-	return err
+	// In case of bufio.Buffer, it will be 0 after io.Copy.
+	literalLen := int64(l.Len())
+
+	n, err := io.Copy(w, l)
+	if err != nil {
+		return err
+	}
+	if n != literalLen {
+		return fmt.Errorf("imap: size of Literal is not equal to Len() (%d != %d)", n, l.Len())
+	}
+
+	return nil
 }
 
 func (w *Writer) writeField(field interface{}) error {
--- a/write.go
+++ b/write.go
@@ -154,7 +154,7 @@ func (w *Writer) writeLiteral(l Literal) error {
 	// In case of bufio.Buffer, it will be 0 after io.Copy.
 	literalLen := int64(l.Len())
 
-	n, err := io.Copy(w, l)
+	n, err := io.CopyN(w, l, literalLen)
 	if err != nil {
 		return err
 	}


write.go
```diff
@@ -151,8 +151,18 @@ func (w *Writer) writeLiteral(l Literal) error {
 		}
 	}
 
-	_, err := io.Copy(w, l)
-	return err
+	// In case of bufio.Buffer, it will be 0 after io.Copy.
+	literalLen := int64(l.Len())
+
+	n, err := io.Copy(w, l)
```
emersion: Maybe we can use `io.CopyN` here?

foxcpp: Oh, right, we should.

write.go
```diff
@@ -151,8 +151,18 @@ func (w *Writer) writeLiteral(l Literal) error {
...
-	_, err := io.Copy(w, l)
-	return err
+	// In case of bufio.Buffer, it will be 0 after io.Copy.
+	literalLen := int64(l.Len())
+
+	n, err := io.CopyN(w, l, literalLen)
+	if err != nil {
+		return err
+	}
+	if n != literalLen {
```
emersion: `n == literalLen` is guaranteed by `io.CopyN`

foxcpp: Not exactly, more like n <= literalLen.
```go
	foo := bytes.Buffer{}
	io.WriteString(&foo, "DIEEEEEEEEEEEEEE")

	fmt.Println(io.CopyN(ioutil.Discard, &foo, int64(len("DIEEEE"))))
	fmt.Println(foo.Len())
```
Will print `6 <nil>` and `10`.

foxcpp: Is it fine for us that Literal may actually contain more data than Len() bytes?

emersion: I don't think so, users will expect the whole `io.Reader` to be consumed. Users can always create an `io.LimitedReader` if needed.

foxcpp:
```
	foo := bytes.Buffer{}
	io.WriteString(&foo, "DIEEEEEEEEEEEEEE")
	literalLen := int64(len("DIEEEE")) // let's say foo.Len() returned this

	fmt.Println(io.CopyN(ioutil.Discard, &foo, literalLen))
	fmt.Println(foo.Len()) // some data is still left in buffer, will print 10
```

emersion: Hmm? `io.CopyN` will return an error anyway right?

foxcpp: It does not care about the actual amount of data in the reader, CopyN only guarantees that it will copy exactly N bytes passed in argument if there is more data than N - it will not return an error.

emersion: Gah, yes, sorry I got confused!

emersion:
Thanks!

PR #251, opened 2019-05-21
Change order of defer statements in tests

foxcpp:
It removes flood of "use of closed connection" errors from stderr.


emersion:
Thanks!

PR #254, opened 2019-05-24
backendutil: Compare header field names in case-insensitive way

foxcpp:
See https://github.com/emersion/maddy/issues/73.


emersion:
Thanks!

PR #256, opened 2019-05-29
backendutil: Use textproto directly

foxcpp:


foxcpp:
I think other backendutil functions should be updated too to use textproto.Header and raw io.Reader instead of message.Entity. Wdyt?

emersion:
>I think other backendutil functions should be updated too to use textproto.Header and raw io.Reader instead of message.Entity. Wdyt?

Yup, that makes sense.

foxcpp:
I'm unsure about FetchEnvelope, shouldn't it see decoded values for From, To, Subject, etc?

emersion:
Thanks!

Issue #175, opened 2018-02-19
Getting raw responses


skinass:
HI.

I want to get raw responses for logging. For example i want to log raw responses from `Login()` and `Authenticate()` even if it is _ok_. How can i get them. And the same question for `Execute()`, can i get raw response in my handler ?

emersion:
You can probably use `Client.SetDebug` for this.

Skinass:
sometimes Client.SetDebug provides incomplete data
For example a small program with only `Client.Login()` and `Client.Logout()` has output
>__C__- W4T8Sg LOGIN testmail@host.com 123123
>__C__- pXeHOg LOGOUT
>__S__- \* BYE IMAP4rev1 Server logging out

when the real tcp exchange looks like
>__S__- \* OK <servname> IMAP4rev1 at imap.host.com:993 ready to talk with ::ffff:185.6.245.226:57620, 2018-Feb-20 15:59:20, KxODiS0qd8d1
>__C__- W4T8Sg LOGIN testmail@host.com 123123
>__S__- \* CAPABILITY IMAP4rev1 CHILDREN UNSELECT LITERAL+ NAMESPACE XLIST BINARY UIDPLUS ENABLE ID IDLE MOVE
>__S__- W4T8Sg OK LOGIN Completed.
>__C__- pXeHOg LOGOUT
>__S__- \* BYE IMAP4rev1 Server logging out
>__S__- pXeHOg OK LOGOUT completed

Skinass:
Found a strange behavior.
As described before. The example bellow won't let you see anything like "123456 OK LOGIN Completed." or any other response from server.
```golang
c , _ := client.New(conn)
c.SetDebug(os.Stdout)
c.Login(user,pswd)
```

BUT.
If you add `c.debug = os.Stdout`  [here](https://github.com/emersion/go-imap/blob/v1/conn.go#L100)
So the debug behaviour gets initialised before the first read from connection, everything works fine. And Stdout provides complete log.

I propose to add a function into "client" package.
It might looks like this:
```golang
func NewWithDebug(conn net.Conn, debug io.Writer) (*Client, error){...}
```
This provides debug setup before  Client initialisation.


However this "bug" looks strange. If i am doing something wrong, please tell me)

emersion:
Fixed in v2.

PR #257, opened 2019-05-30
Fix missing first debug log

mmasztalerczuk:
Hello,
I have a small fix related to logging. Due to implementation, there is impossible to start to debug mode from the beginning. So, I have a proposition to add not required parameter which will make this possible.  
There is an issue for that - https://github.com/emersion/go-imap/issues/175. 

What do you think about it?

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -7,6 +7,7 @@ package client
 
 import (
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -564,7 +565,7 @@ func (c *Client) SetDebug(w io.Writer) {
 }
 
 // New creates a new client from an existing connection.
-func New(conn net.Conn) (*Client, error) {
+func New(conn net.Conn, args ...interface{}) (*Client, error) {
 	continues := make(chan bool)
 	w := imap.NewClientWriter(nil, continues)
 	r := imap.NewReader(nil)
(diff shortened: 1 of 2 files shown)


foxcpp:
IMO that's a weird implementation that loses type safety and makes code much less self-documenting.
If I wanted to introduce a lot of optional arguments that should be set early - I would prefer to have New and NewWithOpts functions with the latter accepting Opts structure with all arguments.
In way similar to how I did it in go-imap-sql: https://godoc.org/github.com/foxcpp/go-imap-sql#New

mmasztalerczuk:
Yep, my implementation has a problem with type safety. 

I was thinking about something similar to your proposition, but in this proposition, I need to implement a new version of DialWithDialerTLS,  DialTLS, DialWithDialer, Dialer, and New. That is a lot of new functions. 

If this is ok, then I can change my proposition for something like that.

foxcpp:
>DialWithDialerTLS, DialTLS, DialWithDialer, Dial~~er~~

DialWithDialer can be dropped in favor of Opts.Dialer. Or left as is if we consider the whole Dial family as a "convenience wrappers" and say that people who need more control should use New directly.

mmasztalerczuk:
Ok, that makes sense. I will create an only new version of New function.

emersion:
Fixed in go-imap v2!

PR #258, opened 2019-06-01
Server-side (reader) support for non-synchronizing literals

foxcpp:
Not sure if we want something for the client side.

### DIFF
--- a/read.go
+++ b/read.go
@@ -234,6 +234,12 @@ func (r *Reader) ReadLiteral() (Literal, error) {
 		return nil, err
 	}
 	lstr = trimSuffix(lstr, literalEnd)
+
+	nonSync := lstr[len(lstr)-1] == '+'
+	if nonSync {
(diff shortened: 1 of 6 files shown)


read.go
```diff
@@ -234,6 +234,12 @@ func (r *Reader) ReadLiteral() (Literal, error) {
 		return nil, err
 	}
 	lstr = trimSuffix(lstr, literalEnd)
+
+	nonSync := lstr[len(lstr)-1] == '+'
```
emersion: This doesn't check whether the string is non-empty. We could use `strings.HasSuffix` to prevent issues like this.

foxcpp: Oops.

emersion:
>Not sure if we want something for the client side.

Client-side support is probably desirable. Can be left as a follow-up issue though.

foxcpp:
Aw, don't write code while being distracted!

emersion:
Thanks!

PR #262, opened 2019-06-06
Client & server support for SASL-IR extension

### LINKED ISSUE #14: Implement SASL-IR
See https://tools.ietf.org/html/rfc4959

foxcpp:
Closes #14.

### DIFF
--- a/client/cmd_noauth.go
+++ b/client/cmd_noauth.go
@@ -111,6 +111,9 @@ func (c *Client) Authenticate(auth sasl.Client) error {
 		InitialResponse: ir,
 		RepliesCh:       make(chan []byte, 10),
 	}
+	if irOk {
+		res.InitialResponse = nil
+	}
 
 	status, err := c.execute(cmd, res)
 	if err != nil {
--- a/client/cmd_noauth.go
+++ b/client/cmd_noauth.go
@@ -2,7 +2,6 @@ package client
 
 import (
 	"crypto/tls"
-	"encoding/base64"
 	"errors"
 	"net"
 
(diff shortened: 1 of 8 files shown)


commands/authenticate.go
```diff
@@ -22,10 +22,17 @@ type AuthenticateConn interface {
 // Authenticate is an AUTHENTICATE command, as defined in RFC 3501 section
 // 6.2.2.
 type Authenticate struct {
-	Mechanism string
+	Mechanism       string
+	InitialResponse string
```
emersion: Can we change this to be a `[]byte`, and encode to base64 in `Command`?

client/cmd_noauth.go
```diff
@@ -98,6 +99,14 @@ func (c *Client) Authenticate(auth sasl.Client) error {
 		Mechanism: mech,
 	}
 
+	irOk, err := c.Support("SASL-IR")
+	if err != nil {
+		return err
+	}
+	if irOk {
+		cmd.InitialResponse = base64.StdEncoding.EncodeToString(ir)
```
emersion: We should probably unset `ir` here, so that it's not duplicated in `responses.Authenticate`?

commands/authenticate.go
foxcpp: Sure, 2766c70.

client/cmd_noauth.go
foxcpp: >responses.Authenticate

You mean commands.Authenticate, right?

foxcpp: What is duplicated tho? I'm not sure if this still applies after two commits I just pushed.

emersion: I'm concerned about `ir` being set for both the command and the response:

```go
	if irOk {
		cmd.InitialResponse = ir // set here
	}

	res := &responses.Authenticate{
		Mechanism:       auth,
		InitialResponse: ir, // set here
		RepliesCh:       make(chan []byte, 10),
	}
```

If we set it for the command, we shouldn't need to set it for the response?

foxcpp: Ah, I see. f22f246.

emersion:
Thanks!

PR #264, opened 2019-06-06
server: Add per-connection panic handler

### LINKED ISSUE #259: Per-connection panic handler
It's annoying that a panic in a connection handler brings down the whole server. We should probably add a panic handler in the connection's goroutine, just like `net/http`.

foxcpp:
Closes #259.

### DIFF
--- a/server/conn.go
+++ b/server/conn.go
@@ -3,8 +3,10 @@ package server
 import (
 	"crypto/tls"
 	"errors"
+	"fmt"
 	"io"
 	"net"
+	"runtime/debug"
 	"time"
 
 	"github.com/emersion/go-imap"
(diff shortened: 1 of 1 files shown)


server/conn.go
```diff
@@ -262,14 +264,28 @@ func (c *conn) silent() *bool {
...
+		if err := recover(); err != nil {
+			c.WriteResp(&imap.StatusResp{
+				Type: imap.StatusRespBye,
+				Info: "Internal server error, closing connection.",
+			})
+
+			stack := debug.Stack()
+			c.s.ErrorLog.Printf("panic serving %v: %v\n%s", c.Info().RemoteAddr, err, stack)
+
+			err = fmt.Errorf("%v", err)
```
emersion: This won't set the outer error, the definition in the `if` shadows the outer one

foxcpp: Oh, right.

PR #266, opened 2019-06-08
Revert "backendutil: Use textproto directly for Match"

foxcpp:
From https://tools.ietf.org/html/rfc3501#page-50:
>The OPTIONAL [CHARSET] specification consists of the word
>"CHARSET" followed by a registered [CHARSET].  It indicates the
>[CHARSET] of the strings that appear in the search criteria.
>[MIME-IMB] content transfer encodings, **and [MIME-HDRS] strings in
>[RFC-2822]/[MIME-IMB] headers, MUST be decoded before comparing
>text in a [CHARSET] other than US-ASCII.**  US-ASCII MUST be
>supported; other [CHARSET]s MAY be supported.

### DIFF
--- a/backend/backendutil/search.go
+++ b/backend/backendutil/search.go
@@ -4,33 +4,33 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"net/mail"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


emersion:
Thanks!

PR #270, opened 2019-06-09
backendutil: Decode headers in Match before comparsion

### LINKED ISSUE #269: backendutil: Match doesn't seem to decode headers
So I was extending go-imap-backend-tests  to test for header/body decoding behavior (#256, #266) when I noticed that backendutil.Match doesn't seem to decode header fields before comparing them. Am I missing something? go-message should decode it, right? 

<details>
<summary>Test case for backendutil</summary>

```go
func TestMatchEncoded(t *testing.T) {
	encodedTestMsg := `From: "fox.cpp" <foxcpp@foxcpp.dev>
To: "fox.cpp" <foxcpp@foxcpp.dev>
Subject: =?utf-8?B?0J/RgNC+0LLQtdGA0LrQsCE=?=
Date: Sun, 09 Jun 2019 00:06:43 +0300
MIME-Version: 1.0
Message-ID: <a2aeb99e-52dd-40d3-b99f-1fdaad77ed98@foxcpp.dev>
Content-Type: text/plain; charset=utf-8; format=flowed
Content-Transfer-Encoding: quoted-printable

=D0=AD=D1=82=D0=BE=D1=82 =D1=82=D0=B5=D0=BA=D1=81=D1=82 =D0=B4=D0=BE=D0=BB=
=D0=B6=D0=B5=D0=BD =D0=B1=D1=8B=D1=82=D1=8C =D0=B7=D0=B0=D0=BA=D0=BE=D0=B4=
=D0=B8=D1=80=D0=BE=D0=B2=D0=B0=D0=BD =D0=B2 base64 =D0=B8=D0=BB=D0=B8 quote=
d-encoding.`
	e, err := message.Read(strings.NewReader(encodedTestMsg))
	if err != nil {
		t.Fatal("Expected no error while reading entity, got:", err)
	}

	crit := imap.SearchCriteria{
		Header: textproto.MIMEHeader{"Subject": []string{"Проверка!"}},
	}

	ok, err := Match(e, 0, 0, time.Now(), []string{}, &crit)
	if err != nil {
		t.Fatal("Expected no error while matching entity, got:", err)
	}

	if !ok {
		t.Error("Expected match for encoded header")
	}
}
```

</details>

foxcpp:
After checking [RFC 2047](https://tools.ietf.org/html/rfc2047) I came to the conclusion that encoded values are allowed almost everywhere. Does it worth it checking whether a particular field is allowed to contain encoded values?

I also added a slightly extended test case I posted with the issue.
Closes #269.

### DIFF
--- a/backend/backendutil/search.go
+++ b/backend/backendutil/search.go
@@ -86,10 +86,7 @@ func Match(e *message.Entity, seqNum, uid uint32, date time.Time, flags []string
 				ok := false
 				values := e.Header.FieldsByKey(key)
 				for values.Next() {
-					decoded, err := values.Text()
-					if err != nil {
-						continue
-					}
+					decoded, _ := values.Text()
 					if matchString(decoded, wantValue) {
 						ok = true
 						break
--- a/backend/backendutil/search.go
+++ b/backend/backendutil/search.go
@@ -86,7 +86,11 @@ func Match(e *message.Entity, seqNum, uid uint32, date time.Time, flags []string
 				ok := false
 				values := e.Header.FieldsByKey(key)
 				for values.Next() {
-					if matchString(values.Value(), wantValue) {
+					decoded, err := values.Text()
+					if err != nil {
+						continue
+					}
+					if matchString(decoded, wantValue) {
 						ok = true
 						break
 					}
--- a/backend/backendutil/search_test.go
+++ b/backend/backendutil/search_test.go
@@ -312,3 +312,51 @@ func TestMatch(t *testing.T) {
 		}
 	}
 }
+
+func TestMatchEncoded(t *testing.T) {
+	encodedTestMsg := `From: "fox.cpp" <foxcpp@foxcpp.dev>
+To: "fox.cpp" <foxcpp@foxcpp.dev>
+Subject: =?utf-8?B?0J/RgNC+0LLQtdGA0LrQsCE=?=
+Date: Sun, 09 Jun 2019 00:06:43 +0300
+MIME-Version: 1.0
+Message-ID: <a2aeb99e-52dd-40d3-b99f-1fdaad77ed98@foxcpp.dev>
+Content-Type: text/plain; charset=utf-8; format=flowed
+Content-Transfer-Encoding: quoted-printable
+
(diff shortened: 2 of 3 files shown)


emersion:
>I came to the conclusion that encoded values are allowed almost everywhere. 

Do you think we should encode/decode *all* fields automatically in go-message?

backend/backendutil/search.go
```diff
@@ -86,7 +86,11 @@ func Match(e *message.Entity, seqNum, uid uint32, date time.Time, flags []string
 				ok := false
 				values := e.Header.FieldsByKey(key)
 				for values.Next() {
-					if matchString(values.Value(), wantValue) {
+					decoded, err := values.Text()
+					if err != nil {
```
emersion: On error (e.g. unknown charset) the raw value is returned. Maybe we should still `matchString` it?

foxcpp: so
```go
decoded, _ := values.Text()
if matchString(decoded, wantValue) {
  ok = true
  break
}
```

I thought it does not return a meaningful value if error is not nil.

emersion: Yeah. I thought I documented it, but it was probably for body charset errors.

foxcpp:
>Do you think we should encode/decode all fields automatically in go-message?

Machine-generated fields generally don't use syntax that can be misinterpreted as an encoded token. User-specified fields allow the use of encoded tokens. Why bother?

emersion:
I have 2 issues with that:

1. We still need to be careful when setting header fields though: some fields (e.g. `To`) become broken if we blindly encode the whole thing (we need to encode only plaintext parts, e.g. `me` in `"me" <me@example.org>`). [I have seen bugs](https://todo.sr.ht/~sircmpwn/aerc2/156) because people were doing `h.SetText("From", "me <me@example.org>")`.
2. Should we silently ignore charset errors?

foxcpp:
>I have 2 issues with that:

Eh, alright, that is more problematic than I thought at first. I guess it is fine to leave go-message as is. For purposes of this backend utility - decoding everything works fine and should not lead to problems.

PR #271, opened 2019-06-09
backendutil: Set Sender and Reply-To to From if empty

### LINKED ISSUE #268: backendutil: Sender and ReplyTo should be set to From if missing
From [RFC 3501, Page 78](https://tools.ietf.org/html/rfc3501#page-78):
>If the Sender or Reply-To lines are absent in the [RFC-2822]
>header, or are present but empty, the server sets the
>corresponding member of the envelope to be the same value as
>the from member (the client is not expected to know to do
>this).

Doesn't seem like backendutil.FetchEnvelope is going it:
https://github.com/emersion/go-imap/blob/00bb0063984c32414ffd9e0117a21ea8f3eefeef/backend/backendutil/envelope_test.go#L17-L18

foxcpp:
Closes #268.


emersion:
Thanks!

PR #273, opened 2019-06-10
client: Use non-synchronizing literals when length is small (<4096)

### LINKED ISSUE #7: Implement IMAP4 non-synchronizing literals
~~See https://tools.ietf.org/html/rfc2088~~

foxcpp:
Closes #7.

### DIFF
--- a/write.go
+++ b/write.go
@@ -49,7 +49,7 @@ func isAscii(s string) bool {
 type Writer struct {
 	io.Writer
 
-	AllowUnsyncLiterals bool
+	AllowAsyncLiterals bool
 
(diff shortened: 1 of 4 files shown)


write.go
```diff
@@ -49,6 +49,8 @@ func isAscii(s string) bool {
 type Writer struct {
 	io.Writer
 
+	AllowUnsyncLiterals bool
```
emersion: "async" is generally used instead of "unsync"

foxcpp: ok

emersion:
Thanks!

PR #274, opened 2019-06-11
backendutil: Handle BODY[1] for non-multipart messages

foxcpp:
This is partial implementation since BODY[something.1] should be also handled for nested messages, but FetchBodySection doesn't seem to support nested messages at all, so let's leave it for another time.


emersion:
Thanks!

PR #280, opened 2019-06-17
backendutil: fix header parsing in FetchEnvelope

### LINKED ISSUE #279: backendutil: FetchEnvelope doen't decode headers too
`FetchEnvelope` uses `headerAddressList` to parse address lists.  
`headerAddressList` tries to use `ParseAddressList` from `net/mail`, but this function knows almost nothing about charsets. So in cases like this one

```From: =?ISO-8859-15?B?TWl0c3VoYSBNaXlhbWl6dQ==?= <mitsuha.miyamizu@example.org>```

the result is just an empty slice, and there's no way to recover the information

### PR DESCRIPTION
**Pull request by cnt0**
this PR fixes #279  by modifying `headerAddressList`. Just decode the value before feeding it to `ParseAddressList`.  

Maybe we should decode all other fields like `Subject` too?

### DIFF
--- a/backend/backendutil/envelope.go
+++ b/backend/backendutil/envelope.go
@@ -5,11 +5,16 @@ import (
 	"strings"
 
 	"github.com/emersion/go-imap"
+	"github.com/emersion/go-message/charset"
 	"github.com/emersion/go-message/textproto"
 )
 
 func headerAddressList(value string) ([]*imap.Address, error) {
-	addrs, err := mail.ParseAddressList(value)
+	decodedValue, err := charset.DecodeHeader(value)
+	if err != nil {
+		return []*imap.Address{}, err
+	}
+	addrs, err := mail.ParseAddressList(decodedValue)
 	if err != nil {
 		return []*imap.Address{}, err
 	}
--- a/backend/backendutil/envelope.go
+++ b/backend/backendutil/envelope.go
@@ -1,20 +1,16 @@
 package backendutil
 
 import (
+	"mime"
 	"net/mail"
 	"strings"
 
 	"github.com/emersion/go-imap"
-	"github.com/emersion/go-message/charset"
 	"github.com/emersion/go-message/textproto"
 )
 
-func headerAddressList(value string) ([]*imap.Address, error) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: backend/backendutil/envelope.go [hunk 1]
```diff
@@ -5,11 +5,16 @@ import (
 	"strings"
 
 	"github.com/emersion/go-imap"
+	"github.com/emersion/go-message/charset"
```
emersion: I'd like to make the dependency to `go-message/charset` optional. Fortunately we already have `imap.CharsetReader` for this purpose.

backend/backendutil/envelope.go
```diff
@@ -5,11 +5,16 @@ import (
...
 	"github.com/emersion/go-message/textproto"
 )
 
 func headerAddressList(value string) ([]*imap.Address, error) {
-	addrs, err := mail.ParseAddressList(value)
+	decodedValue, err := charset.DecodeHeader(value)
```
emersion: It'd be better to use a `mail.AddressParser` instead of double-decoding the header and hope it just works: the decoded value might contain quoted-printable escape sequences.

foxcpp:
Should ENVELOPE contain decode fields tho? 

RFC 3501 is not very clear about it. Relevant: #256.

Dovecot returns encoded fields:
```
* 2 FETCH (ENVELOPE ("Tue, 11 Jun 2019 17:54:30 +0000" {77}
=?utf-8?B?VVRGLTgg0LIg0L/QvtC70LUg0LfQsNCz0L7Qu9C+0LLQug==?= =?utf-8?B?0LA=?= (("Max Mazurov" NIL "fox.cpp" "disroot.org")) (("Max Mazurov" NIL "fox.cpp" "disroot.org")) (("Max Mazurov" NIL "fox.cpp" "disroot.org")) ((NIL NIL "foxcpp" "foxcpp.dev")) NIL NIL NIL "<5a6b59aa38e1c9e1c1393e40273553f0@disroot.org>"))
```

emersion:
Oh right that's a good point. If we don't want to decode them we probably shouldn't use `net/mail` here (or find a way to completely disable the quoted-printable parser).

emersion:

PR #284, opened 2019-06-28
Add date format of current Javamail, as seen in the wild

quite:


emersion:
This is not in the RFC.

I'd like to see a patch for Go first: https://github.com/golang/go/blob/master/src/net/mail/message.go#L74

quite:
You're right of course, and I was overly eager and stopped reading... But it seems Go is already on it, issue: https://golang.org/issue/22661 and change https://go-review.googlesource.com/c/go/+/117596/

And after reading there, it does seem like it is in the RFC:
```
   date-time       =   [ day-of-week "," ] date time [CFWS]
```

emersion:
Sorry I kind of completely forgot about this PR.

>https://go-review.googlesource.com/c/go/+/117596/

Hmm, I think I'm missing how this Go patch is related to the date format added by this pull request.

>Meaning approximately anything (well, limited US-ASCII and so) in parenthesis can go after date time.

Hmm, I don't see any parentheses in the EBNF. Are you referring to header comments?

emersion:
Ohh, now I get it. Disregard my previous comment.

quite:
Hm, how can it be made more robust? I'd say the parsing the affected function does is getting more robust, by grokking another date format seen around...

emersion:
We should remove any trailing CFWS, just like in the Go patch.

quite:
Would removing any trailing `[[:space:]]\+\(.*\)$` cut it, or is that too blunt?

emersion:
Something like that could probably be enough as a first step. A TODO to improve it (support multiple comments and implement more closely the RFC) could be added as a comment.

quite:
Redone like so.

date.go
```diff
@@ -35,33 +36,30 @@ var envelopeDateTimeLayouts = [...]string{
...
-	"Mon, _2 Jan 06 15:04 -0700 (MST)",
 }
 
 // Try parsing the date based on the layouts defined in RFC 5322, section 3.3.
 // Inspired by https://github.com/golang/go/blob/master/src/net/mail/message.go
 func parseMessageDateTime(maybeDate string) (time.Time, error) {
+	// TODO: this is a blunt way to strip any trailing CFWS. A sharper one
+	// would strip multiple comments, and only CFWSs that are really valid
+	// according to RFC5322.
+	re := regexp.MustCompile(`([[:space:]]+\(.*\))$`)
```
emersion: Can we compile outside of the function?

date.go
```diff
@@ -35,33 +36,30 @@ var envelopeDateTimeLayouts = [...]string{
...
 	"Mon, _2 Jan 06 15:04:05 MST",
-	"Mon, _2 Jan 06 15:04:05 -0700 (MST)",
 	"Mon, _2 Jan 06 15:04 -0700",
 	"Mon, _2 Jan 06 15:04 MST",
-	"Mon, _2 Jan 06 15:04 -0700 (MST)",
 }
 
+// TODO: this is a blunt way to strip any trailing CFWS (comments). A sharper
+// one would strip CFWS, and only those really valid according to RFC5322.
+var commentRE = regexp.MustCompile(`([ \t]+\(.*\))$`)
```
emersion: Hmm, you just added some parentheses, is this necessary?

PR #288, opened 2019-08-18
fix detection of multipart in different case

danielkucera:
https://stackoverflow.com/a/9842589/1316157


foxcpp:
https://tools.ietf.org/html/rfc2045#section-5.1
```
     content := "Content-Type" ":" type "/" subtype
                *(";" parameter)
                ; Matching of media type and subtype
                ; is ALWAYS case-insensitive.
```

danielkucera:
@foxcpp , that's a quote from RFC822. Later (p. 13) it states:
```
The type, subtype, and parameter names are not case sensitive.  For
   example, TEXT, Text, and TeXt are all equivalent top-level media
   types.
```

foxcpp:
Thanks!

PR #296, opened 2019-09-15
Make SEARCH charset a RawString

greenfoo:
Right now, the SEARCH command sends something like this:

>  PMn5AW UID SEARCH CHARSET "UTF-8" 1:*

However, those quotes surrounding "UTF-8" seem to be troublesome, at least when talking to the GreenMail server (http://www.icegreen.com/greenmail).

Reading the RFC (https://tools.ietf.org/html/rfc3501#page-49) does not help to understand if those quotes are required or not **but** searching on google for "UID SEARCH CHARSET" seems to return results where quotes are not being used.

This Pull Request removes those quotes (and now GreenMail works as expected).

In any case, feel free to discard this PR if you consider this is something that should be fixed in GreenMail's source code instead of here.

Thanks!

### DIFF
--- a/client/cmd_selected_test.go
+++ b/client/cmd_selected_test.go
@@ -125,7 +125,7 @@ func TestClient_Search(t *testing.T) {
 		done <- err
 	}()
 
-	wantCmd := `SEARCH CHARSET "UTF-8" SINCE "1-Feb-1994" FROM "Smith" DELETED NOT (TO "Pauline")`
+	wantCmd := `SEARCH CHARSET UTF-8 SINCE "1-Feb-1994" FROM "Smith" DELETED NOT (TO "Pauline")`
 	tag, cmd := s.ScanCmd()
 	if cmd != wantCmd {
 		t.Fatalf("client sent command %v, want %v", cmd, wantCmd)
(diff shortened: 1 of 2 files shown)


emersion:
The ABNF is helpful here:

```
search          = "SEARCH" [SP "CHARSET" SP astring] 1*(SP search-key)
```

The charset is an `astring`, which means no quotes indeed. Thus this change LGTM.

Can you update the tests too so that they don't fail?

foxcpp:
>In any case, feel free to discard this PR if you consider this is something that should be fixed in GreenMail's source code instead of here.

`astring` ABNF rule  _allows_ quoted strings, so technically GreenMail is wrong here.

>Reading the RFC (https://tools.ietf.org/html/rfc3501#page-49) does not help to understand if those quotes are required or not but searching on google for "UID SEARCH CHARSET" seems to return results where quotes are not being used.

But I'm afraid many implementations would not understand quoted charset so it might make sense to fix that on our side to improve interoperability.

emersion:
> astring ABNF rule allows quoted strings, so technically GreenMail is wrong here.

Err, indeed. `astring` is either an atom or a string.

greenfoo:
Sorry, this is my first day using "go" and I had no idea how tests worked.
It's fixed now.

PR #299, opened 2019-10-03
Fix search larger and smaller edge case

### LINKED ISSUE #298: backendutil: Match evaluates LARGER using >= instead of >
The code:

```go
if c.Larger > 0 && uint32(n) < c.Larger {
	return false, nil
}
```

should read:

```go
if c.Larger > 0 && uint32(n) <= c.Larger {
	return false, nil
}
```

drauschenbach:
Fixes #298

### DIFF
--- a/backend/backendutil/search.go
+++ b/backend/backendutil/search.go
@@ -116,10 +116,10 @@ func Match(e *message.Entity, seqNum, uid uint32, date time.Time, flags []string
 			return false, err
 		}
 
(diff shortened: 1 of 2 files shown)


foxcpp:
I guess this should be closed in favor of #300.
>Neopallium: I hadn't seen that issue/PR before. PR #299 doesn't include the change to use the whole message size (rfc822 size).

emersion:
Thanks!

Issue #311, opened 2019-11-30
Append possibly returning EOF on reader close


keur:
This issue was discovered in: https://todo.sr.ht/~sircmpwn/aerc2/269

aerc calls Append with an io.Pipe and passes the PipeReader to Append. Users reported a EOF being returned as an error to aerc. I suspect this is caused by the Reader being closed (aerc closes the write immediately after it completes i separate thread) and returning  io.EOF to aerc. If this is the case, the EOF error should not be propagated back to aerc.

foxcpp:
Here is when go-imap client actually reads the literal (io.Reader):
https://github.com/emersion/go-imap/blob/cfc6c581cf592705a35ccc6caf0c7bf73e4f5804/write.go#L154-L160

Relevant part from aerc code is:
```go
func (imapw *IMAPWorker) handleAppendMessage(msg *types.AppendMessage) {
        if err := imapw.client.Append(msg.Destination, msg.Flags, msg.Date,
                &appendLiteral{
                        Reader: msg.Reader,
                        Length: msg.Length,
                }); err != nil {

                imapw.worker.PostMessage(&types.Error{
                        Message: types.RespondTo(msg),
                        Error:   err,
                }, nil)
        } else {
                imapw.worker.PostMessage(&types.Done{types.RespondTo(msg)}, nil)
        }
}
```

I am not sure how that might interfere with PipeReader.Close, does aerc wrap it with the correct msg.Length? I suspect that io.EOF may happen if the actual amount of data sent to the pipe is smaller than the value reported by Len().

foxcpp:
Here is the test case, it fails with Append returning io.EOF. Note the literalWrap.L > len(msg).

```go
func TestClient_Append_Pipe(t *testing.T) {
	c, s := newTestClient(t)
	defer s.Close()

	setClientState(c, imap.AuthenticatedState, nil)

	msg := "Hello World!\r\nHello Gophers!\r\n"
	date := time.Date(2009, time.November, 10, 23, 0, 0, 0, time.UTC)
	flags := []string{imap.SeenFlag, imap.DraftFlag}

	pr, pw := io.Pipe()

	go func() {
		pw.Write([]byte(msg))
		pw.Close()
	}()

	done := make(chan error, 1)
	go func() {
		done <- c.Append("INBOX", flags, date, literalWrap{pr, 35})

		// The buffer is not flushed on error, force it so io.ReadFull can
		// continue.
		c.conn.Flush()
	}()

[...]

PR #312, opened 2019-12-01
Improve handling of mismatched literal length

foxcpp:
See #311.


foxcpp:
I think it could be improved a bit.

Issue #319, opened 2019-12-08
Add BodyStructure.Walk


emersion:
Add `BodyStructure.Walk(fn func(path []int, bs *BodyStructure) (walkChildren bool))` to easily iterate over all parts. The return value decides whether to skip the part.

foxcpp:
It might also be useful to have the ability to stop iteration early. Check how path/filepath.Walk implements this.

emersion:
`path/filepath.Walk`'s implementation is a bit subtle, since `ErrSkip` has a different behaviour when the current file is a directory (don't visit children) and when it's not (don't visit siblings). To completely stop iteration, the only way is a custom error it seems.

I considered using an `error` return value but decided against since `BodyStructure.Walk` itself can't fail internally (that's not the case of `path/filepath.Walk` which can fail because of i/o errors). I'm not sure stopping the iteration completely is necessary as I don't think IMAP messages will ever have more than a few parts.

PR #321, opened 2019-12-13
Introduce BodyStructure.Walk

emersion:
This is a helper to iterate over a body structure tree.

Closes: https://github.com/emersion/go-imap/issues/319

### DIFF
--- a/message.go
+++ b/message.go
@@ -1100,3 +1100,38 @@ func (bs *BodyStructure) Filename() (string, error) {
 	}
 	return decodeHeader(raw)
 }
+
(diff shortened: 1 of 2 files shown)


emersion:
>I think it would be a good idea to document that path argument can be nil for multipart "root". Otherwise it might become the source of unexpected panics when users will try to use it.

How could this trigger a panic? `len(nil)` and `range nil` are legal, and you need to check the length to access an element by index anyway.

foxcpp:
Alright, I guess we can get away saying "assuming that slices have minimal length (1 in this case) unless it is documented is silly so do not do it".

PR #330, opened 2020-01-05
Support ALL criteria in SEARCH

arnisoph:
Signed-off-by: Arnold Bechtoldt <arnold.bechtoldt@inovex.de>

```
criteria := imap.NewSearchCriteria()
c.UidSearch(criteria) // returns: []  Error in IMAP command UID SEARCH: Missing search parameters (0.001 + 0.000 secs).
```

Current behaviour:

```
Z12hwA EXAMINE INBOX
* FLAGS (\Answered \Flagged \Deleted \Seen \Draft)
* OK [PERMANENTFLAGS ()] Read-only mailbox.
* 6 EXISTS
* 6 RECENT
* OK [UNSEEN 1] First unseen.
* OK [UIDVALIDITY 1578248986] UIDs valid
* OK [UIDNEXT 7] Predicted next UID
Z12hwA OK [READ-ONLY] Examine completed (0.001 + 0.000 secs).
hw_TVQ UID SEARCH CHARSET UTF-8
hw_TVQ BAD Error in IMAP command UID SEARCH: Missing search parameters (0.001 + 0.000 secs).
IximMg LOGOUT
* BYE Logging out
IximMg OK Logout completed (0.001 + 0.000 secs).
```

### DIFF
--- a/search.go
+++ b/search.go
@@ -362,5 +362,10 @@ func (c *SearchCriteria) Format() []interface{} {
 		fields = append(fields, RawString("OR"), or[0].Format(), or[1].Format())
 	}
 
+	// Not a single criteria given, add ALL criteria as fallback
+	if len(fields) == 0 {
+		fields = append(fields, RawString("ALL"))
+	}
+
 	return fields
 }
--- a/client/cmd_selected.go
+++ b/client/cmd_selected.go
@@ -122,7 +122,8 @@ func (c *Client) search(uid bool, criteria *imap.SearchCriteria) (ids []uint32, 
 // match the searching criteria. When multiple keys are specified, the result is
 // the intersection (AND function) of all the messages that match those keys.
 // Criteria must be UTF-8 encoded. See RFC 3501 section 6.4.4 for a list of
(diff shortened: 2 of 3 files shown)


search.go
```diff
@@ -59,6 +59,8 @@ func popSearchField(fields []interface{}) (interface{}, []interface{}, error) {
 // SearchCriteria is a search criteria. A message matches the criteria if and
 // only if it matches each one of its fields.
 type SearchCriteria struct {
+	All bool // Search for ALL messages in mailbox
```
emersion: I don't think it's necessary to have a `All` field, since this is the default if no sequence number set or UID set is specified.

arnisoph: I run UidSearch() with an empty SearchCriteria object. See details in the issue descr. Which default do you mean? I haven't found such a default description in RFC 3501.

emersion: I think we should automatically append `ALL` to the list of fields if empty.

arnisoph: Alright, makes sense. I'll update.

Issue #331, opened 2020-01-07
Can't load messages in win10 native client


GZxiaoBinHuang:
I can't load any messages when test it on the win10 native desktop client. https://www.microsoft.com/en-us/p/mail-and-calendar/9wzdncrfhvqm?activetab=pivot:overviewtab
Here are the sample : 
![image](https://user-images.githubusercontent.com/12824536/71866448-1a2ccb80-3141-11ea-904c-79cf5ec28814.png)

I check the log and found that the client send the request to fetch the content as below;
`a18 uid fetch 1 (rfc822.header body.peek[2])`, it uses the 'fetch rfc822.header' to get the headers data, and response of imap-server  set the **BODY[HEADER]** as the key which is incorrect I think, it should be **RFC822.HEADER**.

And the imap-server response when fetch rfc822.header;
`a fetch 1 rfc822.header`
`* 1 FETCH (BODY[HEADER] {3874}`
`content.....`

The Gmail, Aol, Yahoo response of 'fetch rfc822.header' is 
`a fetch 1 rfc822.header`
`* 1 FETCH (RFC822.HEADER {3874}`
`content.....`

I modify the **resp** function at lines 463 in go-imap/message.go
https://github.com/emersion/go-imap/blob/master/message.go
![image](https://user-images.githubusercontent.com/12824536/71866016-c8377600-313f-11ea-891b-fbdf9ad2a40e.png)

And now the imap-server response when fetch rfc822.header;
`a fetch 1 rfc822.header`
`* 1 FETCH (RFC822.HEADER {3874}`
`content.....`

After that, the message content can show normally.
![image](https://user-images.githubusercontent.com/12824536/71866462-23b63380-3141-11ea-9923-cac5631adacc.png)

emersion:
Hmm, I don't remember the exact rules for this. Can you double-check the same happens with other `FetchRFC822` flags? If that's the case, a pull request would be welcome.

GZxiaoBinHuang:
The [RFC3591-page 54](https://tools.ietf.org/html/rfc3501#page-54)  says that the response of untagged FETCH data should return as `RFC822XXX`
![image](https://user-images.githubusercontent.com/12824536/72034015-d5cf3600-32ce-11ea-8617-dc42da279484.png)

[...]

PR #332, opened 2020-01-09
fix rfc822XXX issue

GZxiaoBinHuang:
Fixes: https://github.com/emersion/go-imap/issues/331
fix the rfc822XXX untag response data

### DIFF
--- a/message.go
+++ b/message.go
@@ -463,7 +463,10 @@ func (section *BodySectionName) resp() *BodySectionName {
 	if len(resp.Partial) == 2 {
 		resp.Partial = []int{resp.Partial[0]}
 	}
-	resp.value = ""
+	// resp.value = ""
+    if !strings.HasPrefix(string(resp.value), string(FetchRFC822)) {
+        resp.value = ""
+    }
 	return &resp
 }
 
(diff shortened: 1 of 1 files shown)


message.go
```diff
@@ -463,7 +463,10 @@ func (section *BodySectionName) resp() *BodySectionName {
 	if len(resp.Partial) == 2 {
 		resp.Partial = []int{resp.Partial[0]}
 	}
-	resp.value = ""
+	// resp.value = ""
+    if !strings.HasPrefix(string(resp.value), string(FetchRFC822)) {
```
emersion: Go code is indented with tabs

message.go
```diff
@@ -463,7 +463,10 @@ func (section *BodySectionName) resp() *BodySectionName {
 	if len(resp.Partial) == 2 {
 		resp.Partial = []int{resp.Partial[0]}
 	}
-	resp.value = ""
+	// resp.value = ""
```
emersion: Can we remove this comment out?

GZxiaoBinHuang: yes, I have pushed the code again.

PR #334, opened 2020-01-18
Added support for Gmail's X-GM-RAW search fields

ahornerr:
Hi there!

This is my first contribution to this library and I will say it's quite powerful!

I dug around this morning looking for ways to add arbitrary raw attributes to the search queries and was unable to do so. After further consideration it seems the best way would be to add native X-GM-RAW support to this library!

The specification can be found here https://developers.google.com/gmail/imap/imap-extensions but simply put, we can add a X-GM-RAW attribute to the search query.

I find this necessary to my use case and I believe others may also find it useful. I've added the correct logic and tests.

Thanks!

### DIFF
--- a/search.go
+++ b/search.go
@@ -71,6 +71,7 @@ type SearchCriteria struct {
 	Header textproto.MIMEHeader // Each header field value is present
 	Body   []string             // Each string is in the body
 	Text   []string             // Each string is in the text (header + body)
+	XGMRaw []string             // Gmail X-GM-RAW fields to be added to the query
 
 	WithFlags    []string // Each flag is present
 	WithoutFlags []string // Each flag is not present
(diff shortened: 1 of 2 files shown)


emersion:
Agreed. I really don't want to support GMail's nonstandard flags upstream.

ahornerr:
@foxcpp @emersion This seems like a reasonable (and more flexible) request. I will make some changes.

ahornerr:
@emersion @foxcpp I've made some changes to generalize the raw search functionality. Originally I had Raw as a `map[string][]interface{}` but this made testing and repeatability tricky as Maps in Go are unsorted. By specifying them as an array we can ensure that they're always sorted in the same order. Let me know what you think!

ahornerr:
Actually I don't think this will work. As is my current approach adds parenthesis around the value which Gmail doesn't interpret as a good command.

```
UID SEARCH CHARSET UTF-8 ... X-GM-RAW ("filename:docx OR filename:xlsx")
BAD Could not parse command
```

I'll have to go back to the drawing board with this and correct my issue.

ahornerr:
I've pushed up a better solution that I'm happy with.

Issue #335, opened 2020-01-27
Updates of messages are not sent to all connections


horejsek:
`listenUpdates` creates channel of messages for `MessageUpdate` (https://github.com/emersion/go-imap/blob/v1/server/server.go#L311) which is problematic when two or more connections should get response because only first one will be able to read channel and for the next one(s) it will be already closed channel and will not get response (https://github.com/emersion/go-imap/blob/master/responses/fetch.go#L39). Expunge has the same problem.

Response should be created for each connection to have always fresh channel or Fetch should be using array instead channels. Probably the first should be adopted.

horejsek:
It's not clear if this is really issue by RFC. Updates have much prominent issues which needs to be solved (like when EXPUNGE can be issued). See https://github.com/emersion/go-imap/pull/336

Issue #323, opened 2019-12-15
Delegate updates dispatching to the backend


foxcpp:
The exact problem is outlined in #318. As an alternative solution (and less hacky and buggy), I want to propose to redesign Backend interface in a way that completely moves update handling to the backend side. This includes EXPUNGE updates buffering and sequence numbers translation. This adds additional flexibility in implementation. For example, a backend using MVCC-enabled database can rely on it to provide consistent messages view for clients. Utilities that simplify implementations of rudimentary/naive backends can be added to backendutil package.

As a starting point, here is the snippet with a bunch of Go interfaces:
```go
type Conn interface {
  SendUpdate(u backend.Update)
}

type Backend interface {
  Login(conn Conn, username, password string) (Session, error)
}

type Session interface {
  OpenMbox(mbox string, readOnly bool) (Mailbox, error)
}

type Mailbox interface {
  ...
  Close() error
}
```

## Subtasks

- [ ] Remove existing updates handling code from go-imap/server
- [ ] Define Conn interface for backend and update Backend interfaces
- [ ] Implement utilities for simple backends

emersion:
* It would be nice to have a list of when we're allowed to send each type of unilateral update
* When should the backend know whether it's a good time to send an unilateral update? (e.g. how should the backend know the client issued a `NOOP` and it needs to flush its unilateral update buffer)
* I see some overlap between `Conn` and `Session`. I wonder if there's a way to simplify this (e.g. with something like `Server.SendUpdate(s Session, u backend.Update)`).
* I'd really like a collection of `backendutils` helpers to handle all of this. I don't expect backend authors to know everything about IMAP.

foxcpp:
>It would be nice to have a list of when we're allowed to send each type of unilateral update

AFAIK, will check RFCs later:
- FLAGS, EXISTS, RECENT can be sent in response to any command with a selected mailbox.
- EXPUNGE cannot be sent in response to pipeline-able commands using sequence numbers (that includes STORE, FETCH but not UID FETCH, UID STORE and COPY, later is not permitted to be pipelined).
- For interoperability purposes, it is a good idea to not send updates between commands (after the tagged response to the previous command, but before the next command is received). We cannot expect client to actually consume them and take into account when sending the next command. 
- Previous statement applies for EXISTS update too. It will affect the effect of `STORE 1:* +FLAGS (\Deleted)`, for example.

[...]

PR #336, opened 2020-02-11
Fix imap idle when multiple connections are present

houlahaj:
If there are multiple connections, we need to send them each the same updates. This currently doesn't work correctly because messages and sequence numbers are held in a channel which is emptied after the first connection's updates are sent out, leaving none for subsequent connections.

We avoid this by just doing the whole process anew for each connection. Perhaps not the best solution but the simplest given that channels are used in many places and switching to something that doesn't empty on first pass (like a slice) would be a much larger change.

Furthermore, updates are sent on the responses channel in serial as opposed to in parallel. This is to ensure order is preserved. Suppose two updates are sent in quick succession: we create two goroutines, each tasked with writing one update to the responses channel, but who knows which is actually executed first, so it's possible the latter update is written to the responses channel first, mixing up the order of imap updates. This change circumvents this issue, though still waits on all updates in separate goroutines so as to not block too much.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -298,34 +298,34 @@ func (s *Server) listenUpdates() {
 		update := <-s.Updates
 
 		var res imap.WriterTo
-		switch update := update.(type) {
-		case *backend.StatusUpdate:
-			res = update.StatusResp
-		case *backend.MailboxUpdate:
-			res = &responses.Select{Mailbox: update.MailboxStatus}
-		case *backend.MessageUpdate:
-			ch := make(chan *imap.Message, 1)
-			ch <- update.Message
-			close(ch)
 
-			res = &responses.Fetch{Messages: ch}
-		case *backend.ExpungeUpdate:
-			ch := make(chan uint32, 1)
-			ch <- update.SeqNum
-			close(ch)
-
-			res = &responses.Expunge{SeqNums: ch}
-		default:
-			s.ErrorLog.Printf("unhandled update: %T\n", update)
-		}
-		if res == nil {
-			continue
-		}
-
-		sends := make(chan struct{})
-		wait := 0
+		wg := sync.WaitGroup{}
 		s.locker.Lock()
 		for conn := range s.conns {
+			switch update := update.(type) {
+			case *backend.StatusUpdate:
+				res = update.StatusResp
+			case *backend.MailboxUpdate:
+				res = &responses.Select{Mailbox: update.MailboxStatus}
+			case *backend.MessageUpdate:
+				ch := make(chan *imap.Message, 1)
+				ch <- update.Message
+				close(ch)
+
+				res = &responses.Fetch{Messages: ch}
+			case *backend.ExpungeUpdate:
+				ch := make(chan uint32, 1)
+				ch <- update.SeqNum
+				close(ch)
+
+				res = &responses.Expunge{SeqNums: ch}
+			default:
+				s.ErrorLog.Printf("unhandled update: %T\n", update)
+			}
+			if res == nil {
+				continue
+			}
+
 			ctx := conn.Context()
 
 			if update.Username() != "" && (ctx.User == nil || ctx.User.Username() != update.Username()) {
@@ -341,32 +341,23 @@ func (s *Server) listenUpdates() {
 				}
 			}
 
-			conn := conn // Copy conn to a local variable
-			go func() {
-				done := make(chan struct{})
-				conn.Context().Responses <- &response{
-					response: res,
-					done:     done,
-				}
-				<-done
-				sends <- struct{}{}
-			}()
+			done := make(chan struct{})
+			conn.Context().Responses <- &response{
+				response: res,
+				done:     done,
+			}
 
-			wait++
+			wg.Add(1)
+			go func() {
+				<-done
+				wg.Done()
+			}()
 		}
 		s.locker.Unlock()
 
-		if wait > 0 {
-			go func() {
-				for done := 0; done < wait; done++ {
-					<-sends
-				}
+		wg.Wait()
 
-				close(update.Done())
-			}()
-		} else {
-			close(update.Done())
-		}
+		close(update.Done())
 	}
 }
 


horejsek:
This is fix for #335 I reported. But now I have serious doubt it's correct. Sending `EXPUNGE` with index more than once can lead to deleting unwanted messages.

* Let's say client is connected with three connection, both selected in same mailbox. Mailbox has 10 messages.
* Two connections are in IDLE or NOOP or other selected command.
* Third one will delete some messages, leading to `EXPUNGE` update with index 1.
* That `EXPUNGE` will go to both connections.
* Client will receive this three times. Once for `EXPUNGE` command and probably client will not delete the message as it was already deleted(?), but it will not ignore it for the other two connections. It will delete again message on the same index.
* Result is mailbox has only 7 or 8 messages (depends how clients treat untagged `EXPUNGE` result in `EXPUNGE` command), not 9.

I'm reading RFC3501 in detail and probably there is other problem, but currently I'm more unsure how it should really work. So I would wait with merging this one.

emersion:
>Client will receive this three times

Do you mean that the client will receive three identical `EXPUNGE` updates, one on each connection?

>but it will not ignore it for the other two connections. It will delete again message on the same index.

I don't think so? Since the updates are sent on different connections it should be fine?

horejsek:
> Do you mean that the client will receive three identical EXPUNGE updates, one on each connection?

Yes.

> I don't think so? Since the updates are sent on different connections it should be fine?

Not really. How client can know which EXPUNGE update processed and which not? Client will process all of them.

From RFC: "The EXPUNGE response reports that the specified message sequence number has been permanently removed from the mailbox." Notice the sequence number. Example of EXPUNGE: `* 44 EXPUNGE`.

If mailbox has 5 messages (sequence to UID: 1: 100, 2: 200, 3: 300, 4: 400, 5: 500) and client will open more connections at the same time, we will respond for example with `* 2 EXPUNGE` twice. After first EXPUNGE client will update mailbox to 1: 100, 2: 300, 3: 400, 4: 500 and then there will be again EXPUNGE update in different connection leading to: 1: 100, 2: 400, 3: 500 and message with UID 300 is wrongly deleted.

By the way there is also limit in RFC:

> An EXPUNGE response MUST NOT be sent when no command is in progress, nor while responding to a FETCH, STORE, or SEARCH command. This rule is necessary to prevent a loss of synchronization of message sequence numbers between client and server.

But actually this MR is not breaking this. This bug is already here, because EXPUNGE doesn't have bug reading channel only once, so it's already now sent to all open connections.

-----

What's actually the problem is respond is not send to correct connection. For example STORE command should return FETCH untagged responses. Now if there is second connection with active IDLE and STORE command should return for example 10 FETCH untagged responses (marking 10 messages as read), it will return some responses as result to STORE command and some to IDLE command because of bug with the channel is read only once. But I have strong feeling now it should return only to STORE command and not to every connection.

From RFC I can see each command has it's own results which should get for sure. So STORE command should be changed to always return it's own FETCH commands, the same was how FETCH is done, instead of using Updates channel which it can send to any open connection.

And processing of Updates should be changed to always use only one connection, not all of them. But it has to be always at least one connection. If no connection is not open, there should be some queue. No update should be simply thrown away. From RFC 2177:

> At that point, the server MAY send any remaining queued untagged responses and then MUST immediately send the tagged response to the IDLE command and prepare to process other commands.

-----

To sum it up, if I understand problem clearly:

* IDLE should not discard updates, there should be queue
* Updates should not be send to all open connections, definitely not EXPUNGE
* STORE should return own FETCH responses instead of using Updates channel
* EXPUNGE should also return own EXPUNGE responses instead of using Updates channel

horejsek:
There is another problem with EXPUNGE: we cannot send EXPUNGE any time. EXPUNGE can be send only to some connection which allows it. It is allowed based on running command. For example it cannot happen for FETCH, STORE and SEARCH (those commands do not permit EXPUNGEs at all), it has to be UID version of FETCH, STORE or SEARCH, because EXPUNGE is invalidating sequence numbers. Also, EXPUNGE cannot happen in permitted connection during ongoing other command which could invalidate sequence numbers. First running command has to be finished, then EXPUNGE needs to be done and not allow to run any other command before it's finished.

So this is another problem with Updates channel which is not handled if I understand the code correctly.

RFC 3501:

```
   A non-obvious ambiguity occurs with commands that permit an untagged
   EXPUNGE response (commands other than FETCH, STORE, and SEARCH),
   since an untagged EXPUNGE response can invalidate sequence numbers in
   a subsequent command.  This is not a problem for FETCH, STORE, or
   SEARCH commands because servers are prohibited from sending EXPUNGE
   responses while any of those commands are in progress.  Therefore, if
   the client sends any command other than FETCH, STORE, or SEARCH, it
   MUST wait for the completion result response before sending a command
   with message sequence numbers.
```

And good explanation in this mailing list: https://dovecot.org/pipermail/dovecot/2017-December/110351.html

horejsek:
About this MR: one change is about sending responses. Originally it was done in goroutine and it works, this changes to make it in blocking way. It's kind of fine with Thundebird for example, but Outlook on Mac for me takes some not reasonably small time (not sure why). If the channel has not enough of space, it panics; if there is safety net with timeout, it will discard many updates because of it. Not sure what can make delay, but I think it shouldn't be changed.

horejsek:
I noticed another bug. :-) Server can send untagged response even if there is no command in progress, but it's not true for EXPUNGE. From RFC:

```
   Server implementations are permitted to send an untagged response
   (except for EXPUNGE) while there is no command in progress.
```

```
      An EXPUNGE response MUST NOT be sent when no command is in
      progress, nor while responding to a FETCH, STORE, or SEARCH
      command.  This rule is necessary to prevent a loss of
      synchronization of message sequence numbers between client and
      server.  A command is not "in progress" until the complete command
      has been received; in particular, a command is not "in progress"
      during the negotiation of command continuation.
```

foxcpp:
>Server can send untagged response even if there is no command in progress, but it's not true for EXPUNGE. Current behaviour is sending any update to any state

It is a rather hard to fix design issue, there is more than just "don't send updates without running commands". See issues assigned for 2.0 milestone.

horejsek:
> See issues assigned for 2.0 milestone.

Ah, sorry, I didn't check if there is already issue for it. 🤦‍♂ 


Anyway, I was going through more how clients work. I can confirm TB, Outlook or Apple Mail (all on Mac) opens only one connection per user and mailbox. It's not forbidden by RFC but seems clients do that. So this is actually not a bug per se. In case we consider each connection is different agent, then it's intended to send the update to all agents (TB and Outlook for example).

> IDLE should not discard updates, there should be queue

This is really confusing, but when I checked communication between Google and TB, for example, Google do not queue updates. By RFC is required server sends FLAGS, EXISTS and RECENT which should be enough to update client's mailbox to latest state. And only then there has to be queue for cases when update cannot be issued right away (for example it's EXPUNGE).

> It is a rather hard to fix design issue

I understand. I see #323 discussing this topic. v2 should then fix issues when to send EXPUNGE properly and STORE.SILENT should not disable all updates.


It's definitely not part of this MR, but still this MR should be improved, because we cannot block channel as proposed here. Outlook takes time and doesn't work it well then.

jameshoulahan:
Closing. This MR is borked (sometimes deadlocks). Issue is real though and needs to be addressed (#367).

Issue #322, opened 2019-12-15
Merge extensions into the main code base


foxcpp:
Comments in #260 outline a number of issues caused by an attempt to keep extensions independent of the main go-imap code. This includes two extensions being unable to extend the same command.
I propose to ditch the idea of external extensions or at least seriously reconsider how it works for go-imap v2. 

There is number of important IMAP extensions, notably IDLE, UIDPLUS and CONDSTORE. I think they should be supported by go-imap itself. This will simply make implementing its support on the server and client sides simpler.

## Subtasks

- [ ] Revise the command parser in accordance with [RFC 4466].
- [ ] Define Server-Backend interface suitable for use with multiple extensions at the same time.
- [ ] Define Client interface suitable for use with multiple extensions at the same time (e.g. UIDPLUS + CONDSTORE).
- [ ] Merge certain (probably all) extensions into the main code base.
or
- [ ] Define new interfaces for use by external extensions.

## Extensions to consider for merge

- IDLE
  [RFC 2177](https://tools.ietf.org/html/rfc2177), [Implementation](https://github.com/emersion/go-imap-idle)
  Rationale: Fundamental for useful clients and servers implementations.
- ENABLE
  [RFC 53161](https://tools.ietf.org/html/rfc5161), [Implementation](https://github.com/emersion/go-imap-enable)
  Rationale: Heavily used by other extensions

- CHILDREN
  [RFC 3348](https://tools.ietf.org/html/rfc3348), [Implementation](https://github.com/foxcpp/go-imap-sql/tree/master/children)
  Rationale: Too small to be kept separate.
- SPECIAL-USE
  [RFC 6154](https://tools.ietf.org/html/rfc6154), [Implementation](https://github.com/emersion/go-imap-specialuse)
  Rationale: Too small to be kept separate.
- UNSELECT
  [RFC 3691](https://tools.ietf.org/html/rfc3691), [Implementation](https://github.com/emersion/go-imap-unselect)
  Rationale: Too small to be kept separate.

... pretty much every small extension, currently maintained in a separate repository, see https://github.com/emersion/go-imap#extensions ...

## References

- Collected Extensions to IMAP4 ABNF
  [RFC 4466](https://tools.ietf.org/html/rfc4466)

[RFC 4466]: https://tools.ietf.org/html/rfc4466

emersion:
Note that this isn't necessarily a breaking change, we can add new symbols as long as we don't break the current API.

arnisoph:
[...]

PR #341, opened 2020-03-04
server: Add Mailbox.Close, make ListMailboxes return MailboxInfo

### LINKED ISSUE #337: Add Mailbox.Close
I am working on the maildir-based backend that uses BoltDB for caches and metadata.
BoltDB DB is created per mailbox. backend.Mailbox implementation opens it on demand, but it also needs to close it at some point. Currently Close() is missing for Mailbox so I cannot do it.

Proposal: If Mailbox implements io.Closer, call Close() after use. Not sure about ListMailboxes.

foxcpp:
~~This change is backward compatible with v1.~~

go-imap-unselect needs an update to avoid 'leaking' mailboxes, I propose to merge it into go-imap along the way per #322.

Closes #337.

### DIFF
--- a/server/cmd_auth.go
+++ b/server/cmd_auth.go
@@ -2,7 +2,6 @@ package server
 
 import (
 	"errors"
-	"io"
 
 	"github.com/emersion/go-imap"
 	"github.com/emersion/go-imap/backend"
@@ -37,12 +36,12 @@ func (cmd *Select) Handle(conn Conn) error {
 
 	status, err := mbox.Status(items)
 	if err != nil {
-		closeMailbox(mbox)
+		mbox.Close()
 		return err
 	}
 
 	if ctx.Mailbox != nil {
-		closeMailbox(ctx.Mailbox)
+		ctx.Mailbox.Close()
 	}
 	ctx.Mailbox = mbox
 	ctx.MailboxReadOnly = cmd.ReadOnly || status.ReadOnly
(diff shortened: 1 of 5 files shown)


emersion:
Honestly I'm not a big fan of this change. This doesn't fit well with the v1 interfaces. I'd prefer to start working on a v2 branch to address this issue (by adding a mandatory `Mailbox.Close` method and making `ListMailboxes` return `MailboxInfo` structs directly).

foxcpp:
Alright, I will update this PR this weekend (hopefully, soju :\ ).

foxcpp:
I guess, https://github.com/foxcpp/go-imap-maildir will be built along with go-imap v2 then :)

server/cmd_auth.go
```diff
@@ -110,6 +114,7 @@ func (cmd *Subscribe) Handle(conn Conn) error {
 	if err != nil {
 		return err
 	}
+	defer mbox.Close()
 
 	return mbox.SetSubscribed(true)
```
emersion, line 119: Maybe this should be a `User` method instead, so that a mailbox is only opened on SELECT?

emersion:
Overall looks good. We could address the comment in another PR if you prefer.

server/cmd_auth.go
foxcpp: I agree. I think I even mentioned that somewhere but this is out-of-scope of this PR.

PR #342, opened 2020-03-04
Merge support for COMPRESS extension

foxcpp:
First round of changes for #322.

- [x] APPENDLIMIT
- [x] COMPRESS
- [x] ENABLE
- [ ] ~~ID~~
- [x] SPECIAL-USE
- [x] UNSELECT
- [x] [CHILDREN](https://github.com/foxcpp/go-imap-sql/tree/master/children)
- [ ] [I18NLEVEL](https://github.com/foxcpp/go-imap-i18nlevel) (perhaps level 1 only?)

This PR does not include server-side change to the UNSELECT command required by #341 (needs rebase after #341 gets merged)

### DIFF
--- a/deflate.go
+++ b/deflate.go
@@ -1,0 +1,17 @@
+package imap
+
+// A CompressUnsuppportedError is returned by Client.Compress when the provided
+// compression mechanism is not supported.
+type CompressUnsupportedError struct {
+	Mechanism string
+}
+
+func (err CompressUnsupportedError) Error() string {
+	return "COMPRESS mechanism " + err.Mechanism + " not supported"
+}
+
+// Compression algorithms for use with COMPRESS extension (RFC 4978).
+const (
+	// The DEFLATE algorithm, defined in RFC 1951.
+	CompressDeflate = "DEFLATE"
+)
--- a/README.md
+++ b/README.md
@@ -135,6 +135,7 @@ includes:
 * [CHILDREN](https://tools.ietf.org/html/rfc3348)
 * [UNSELECT](https://tools.ietf.org/html/rfc3691)
 * [APPENDLIMIT](https://tools.ietf.org/html/rfc7889)
+* [COMPRESS](https://tools.ietf.org/html/rfc4978)
 
 Support for other extensions is provided via separate packages. See below.
 
@@ -146,7 +147,6 @@ Commands defined in IMAP extensions are available in other packages. See [the
 wiki](https://github.com/emersion/go-imap/wiki/Using-extensions#using-client-extensions)
 to learn how to use them.
 
-* [COMPRESS](https://github.com/emersion/go-imap-compress)
 * [ENABLE](https://github.com/emersion/go-imap-enable)
 * [ID](https://github.com/ProtonMail/go-imap-id)
 * [IDLE](https://github.com/emersion/go-imap-idle)
--- a/client/client.go
+++ b/client/client.go
@@ -66,9 +66,10 @@ type Client struct {
 	isTLS      bool
 	serverName string
 
-	loggedOut chan struct{}
-	continues chan<- bool
-	upgrading bool
+	loggedOut    chan struct{}
+	continues    chan<- bool
+	upgrading    bool
+	isCompressed bool
 
 	handlers       []responses.Handler
 	handlersLocker sync.Mutex
--- a/client/cmd_any.go
+++ b/client/cmd_any.go
@@ -1,15 +1,22 @@
 package client
 
 import (
+	"compress/flate"
 	"errors"
+	"net"
 
 	"github.com/emersion/go-imap"
 	"github.com/emersion/go-imap/commands"
+	"github.com/emersion/go-imap/internal"
 )
 
 // ErrAlreadyLoggedOut is returned if Logout is called when the client is
 // already logged out.
 var ErrAlreadyLoggedOut = errors.New("Already logged out")
+
+// ErrAlreadyCompress is returned by Client.Compress when compression has
+// already been enabled on the client.
+var ErrAlreadyCompressed = errors.New("COMPRESS is already enabled")
 
 // Capability requests a listing of capabilities that the server supports.
 // Capabilities are often returned by the server with the greeting or with the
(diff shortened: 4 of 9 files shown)


foxcpp:
On I18NLEVEL=1 vs 1+2 support: Client support for I18NLEVEL altogether is not very impressive according to [DCS](https://codesearch.debian.net/search?q=I18NLEVEL&page=1) and nobody seems to care about I18NLEVEL=2.

foxcpp:
@emersion What's your opinion on ID extension? Not worth merging?

server/cmd_noauth_test.go
```diff
@@ -30,7 +30,7 @@ func testServerTLS(t *testing.T) (s *server.Server, c net.Conn, scanner *bufio.S
 
 	io.WriteString(c, "a001 CAPABILITY\r\n")
 	scanner.Scan()
-	if scanner.Text() != "* CAPABILITY IMAP4rev1 LITERAL+ SASL-IR STARTTLS LOGINDISABLED" {
+	if scanner.Text() != "* CAPABILITY IMAP4rev1 LITERAL+ SASL-IR SPECIAL-USE CHILDREN UNSELECT COMPRESS=DEFLATE APPENDLIMIT STARTTLS LOGINDISABLED" {
```
emersion: Maybe we can add constants for these? So that we only need to update them once when adding a new extension?

emersion:
When this is merged, we need to create issues if the client or server side of an extension is missing.

emersion:
>What's your opinion on ID extension? Not worth merging?

Yeah, I'd prefer to leave it out if possible.

foxcpp:
Not sure what is better for I18NLEVEL. Is just having a variable Server.EnableI18N enough or do we want to check backend I18N "awareness" using some interface like `I18NSupported()`

foxcpp:
Oh, you know, lets get that merged and then I will continue working on it when I will have more time.

backend/appendlimit.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"errors"
+)
+
+// An error that should be returned by User.CreateMessage when the message size
+// is too big.
+var ErrTooBig = errors.New("Message size exceeding limit")
+
+// A backend that supports retrieving per-user message size limits.
+type AppendLimitBack interface {
```
emersion, line 12: Let's keep the name "Backend"

client/cmd_auth.go
```diff
@@ -252,3 +252,30 @@ func (c *Client) Append(mbox string, flags []string, date time.Time, msg imap.Li
...
+// Unselect frees server's resources associated with the selected mailbox and
+// returns the server to the authenticated state. This command performs the same
+// actions as Close, except that no messages are permanently removed from the
+// currently selected mailbox.
+//
+// If client does not support the UNSELECT extension, ErrExtensionUnsupported
+// is returned.
+func (c *Client) Unselect() error {
+	if ok, err := c.Support("UNSELECT"); !ok || err != nil {
+
```
emersion, line 265: Missing `return`?

deflate.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	return c.Conn.Close()
+}
+
+// CreateDeflateConn wraps net.Conn with flate.Writer and flate.Reader.
+//
+// It is used by COMPRESS extension implementation and it not a part of the
+// library API. Do not use it.
+func CreateDeflateConn(c net.Conn, level int) (net.Conn, error) {
```
emersion, line 66: Should we expose this to users? Or should we define this in an `internal` package?

backend/appendlimit.go
emersion, line 12: Should we embed the `Backend` interface here?

(Same goes for `User`)

deflate.go
foxcpp: Not seeing a reason to keep it public.

client/cmd_auth.go
foxcpp: I copied code verbatim from go-imap-unselect. Seems so.

backend/appendlimit.go
foxcpp: Most external extensions currently do not embed it. Though embedding it could be more convenient in some cases.  => We should

foxcpp: https://github.com/emersion/go-imap/pull/342/commits/430faf9d88b20f3609b4aca3830f506bc98ba59b

deflate.go
foxcpp: https://github.com/emersion/go-imap/pull/342/commits/2d8e2f30deecf1b156d11628a700434a88c84afd

client/cmd_auth.go
foxcpp: https://github.com/emersion/go-imap/pull/342/commits/c6380dd230b00638df4d07e949bbff6c4ca65fe9

backend/appendlimit.go
foxcpp: https://github.com/emersion/go-imap/pull/342/commits/430faf9d88b20f3609b4aca3830f506bc98ba59b

emersion:
```
17:39 @emersion foxcpp: sorry if we've already discussed about that, but do we want to unconditionally enable the SPECIAL-USE and CHILDREN extensions?
17:39 @emersion we don't want to advertise it if the backend doesn't properly sets those attrs
```

foxcpp:
Advertising SPECIAL-USE without backend support is valid - it will just look like there MAY be special-use mailboxes but there are currently none. CREATE-SPECIAL-USE is a separate capability. 
CHILDREN may be a problem though.

emersion:
OK that makes sense

foxcpp:
RFC 3348 seems to be pretty lax regarding attributes too:
>   The CHILDREN extension defines two new attributes that MAY be
>   returned within a LIST response.

So we probably can get away with it without bothering much.

emersion:
All right. I rejiggered and pushed these commits:

- 80111546d218b444a22a2965f6b615c64d90d418
- 3e25bca974bb1bf7bfd7746088e2aafb80315607
- 61057f7c97725d9a83c21973d3f2f47ba84b53c9
- 7b7dd371b82ad5423c6072625f5148e541153542

emersion:
Holding off this PR for now because merging it would break the API: some extensions might get added twice (built-in + external).

foxcpp:
What is the effect of adding an external extension when it is already supported directly? Can we add a compatibility hack that would just ignore the external extension in this case?

emersion:
I wonder if it's better to let external extensions overwrite built-in extensions (and de-dup our caps), or make if it's better to make `EnableExtension` a no-op if already enabled (might break the API?).

foxcpp:
>I wonder if it's better to let external extensions overwrite built-in extensions (and de-dup our caps)

The idea with  builtin extensions is that they are builtin because their support requires special changes to go-imap 'core' source code. If we let external extensions override such changes we need to have a whole bunch of `if` all around go-imap to disable corresponding changes and let the external extension do its thing (possibly incorrectly).

>or make if it's better to make EnableExtension a no-op if already enabled (might break the API?)

It _will_ break API with some extensions so some careful judgement is needed to figure out what we can merge into v1. Some breakage may be acceptable, for example merging https://github.com/foxcpp/go-imap-namespace is probably okay - it will prevent backends from advertising custom namespaces but I do not expect this to be used often and it is unlikely to cause any real breakage e.g. for clients.

CrawX:
Even after merging emersion/go-imap-compress#6, COMPRESS doesn't work for me because the reader thread reads compressed data without the deflater in place which messes up the connection state completely. I found a workaround (see [this](/CrawX/go-imap/tree/fix-compress-upgrade-race) and [this](/CrawX/go-imap-compress/tree/fix-compress-upgrade-race). With these two additional changes, COMPRESS seems to actually work for me :) Does it make sense to try to get them merged? I feel like #342 might also address this race condition issue.

In any case, please include the raw-string fix from emersion/go-imap-compress#6.

emersion:
OK, so for v1, you think it's best to ignore `EnableExtension` calls if we already have that extension enabled? This sounds like a reasonable workaround.

emersion:
@CrawX That's weird because e.g. STARTTLS works correctly... In any case it's unrelated to this PR, so please open a separate issue about it.

emersion:
Merged UNSELECT and APPENDLIMIT. Not sure COMPRESS is worth the complexity…

emersion:
Superseded by https://github.com/emersion/go-imap/pull/606

PR #350, opened 2020-04-08
Add MailboxInfoUpdate

horejsek:
By RFC any untagged response can be send from server anytime. INFO is useful to notify client that new mailbox was created on server. This adds only support for this update but for v2 would be good to have ability to send update of any kind. Probably `Update` interface could contain function to generate response and even if go-imap do not support some update any app could create add without need to change `listenUpdates` in go-imap server.

### DIFF
--- a/backend/updates.go
+++ b/backend/updates.go
@@ -60,6 +60,12 @@ type MailboxUpdate struct {
 	*imap.MailboxStatus
 }
 
+// MailboxInfoUpdate is a maiblox info update.
+type MailboxInfoUpdate struct {
+	Update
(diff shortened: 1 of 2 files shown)


emersion:
>INFO is useful to notify client that new mailbox was created on server

Is this something e.g. Dovecot does?

horejsek:
> Is this something e.g. Dovecot does?

I don't know Dovecot. But this is something what is allowed by RFC and only way how to notify client about new mailbox. I tested this and for example Apple Mail, TB or Outlook are really responding to this correctly.

PR #358, opened 2020-05-08
server: Terminate SASL exchange with proper response on "*"

foxcpp:


status.go
```diff
@@ -118,3 +118,16 @@ func (r *StatusResp) WriteTo(w *Writer) error {
 
 	return w.writeCrlf()
 }
+
+// ErrStatusResp can be returned by a Handler to replace the default status
```
emersion, line 122: Handler is a server thing. Maybe we can document that `imap.ErrStatusResp` can be used in `server.Handler` instead?

status.go
```diff
@@ -118,3 +118,16 @@ func (r *StatusResp) WriteTo(w *Writer) error {
...
+// response. The response tag must be empty.
+//
+// To suppress default response, Resp should be set to nil.
+type ErrStatusResp struct {
+	// Response to send instead of default.
+	Resp *StatusResp
+}
+
+func (err *ErrStatusResp) Error() string {
+	return ""
```
emersion, line 132: Unrelated to this PR, but this is a little bit weird. Maybe we should change this.

status.go
foxcpp: server.Handler.Handle code already references ErrStatusResp. I updated it to reference imap.ErrStatusResp and also fixed ErrStatusResp doc to reference server.Handler. Is that better?

status.go
foxcpp: I am not entirely sure about contexts where StatusResp is used.
Would this work fine?
```go
func (err *ErrStatusResp) Error() string {
	if err.Resp == nil {
		return "imap: suppressed response"
	}
	return err.Resp.Info
}
```

emersion: Yeah, this looks good to me. It's just to avoid surprising and hard-to-debug error messages.

status.go
emersion: Looks good!

emersion:
Can you split the changes in logical commits, removing the fixup commits?

foxcpp:
> Can you split the changes in logical commits, removing the fixup commits?

Done.

PR #359, opened 2020-05-21
readme: use variable instead of mbox.Message

idirouhab:
It's assigning mbox.Messages to `to` but then in the if condition is not used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ func main() {
 	// Get the last 4 messages
 	from := uint32(1)
 	to := mbox.Messages
-	if mbox.Messages > 3 {
+	if to > 3 {
 		// We're using unsigned integers here, only substract if the result is > 0
 		from = mbox.Messages - 3
 	}


README.md
```diff
@@ -70,7 +70,7 @@ func main() {
 	// Get the last 4 messages
 	from := uint32(1)
 	to := mbox.Messages
-	if mbox.Messages > 3 {
+	if to > 3 {
```
emersion, line 73: This condition is checking whether `mbox.Messages - 3 > 0` (see the comment). I think it makes more sense to keep it as-is.

idirouhab: yes, the if the condition it does make a lot of sense. My suggestion is not regarding the implementation but the usage of a variable that it was assigned just right before the if:


to has already mbox.Messages assigned that's why I thought it would make sense to use it in the if condition.

emersion: The variable is used after, in the `AddRange` call.

PR #362, opened 2020-06-01
fix: empty envelope address fields

### LINKED ISSUE #361: Envelope sender should be `nil` instead of `()`
As stated in [this issue](https://github.com/jonhoo/rust-imap/issues/166), the envelope parsing does not adhere to the spec.

It might be an issue i should fill in `go-imap` but i'm not sure about that, so i post it here in the meantime.

kyazdani42:
fixes #361.


message.go
```diff
@@ -707,7 +707,12 @@ func ParseAddressList(fields []interface{}) (addrs []*Address) {
 
 // Format an address list to fields.
 func FormatAddressList(addrs []*Address) (fields []interface{}) {
-	fields = make([]interface{}, len(addrs))
+	addr_len := len(addrs)
```
emersion, line 710: Style: Go uses `camelCase`. But I don't think it's worth it to define a variable here, we can just inline `len(addrs)` at both sites.

message_test.go
```diff
@@ -391,6 +391,15 @@ func TestAddressList(t *testing.T) {
 	}
 }
 
+func TestEmptyAddressList(t *testing.T) {
+    addrs := make([]*Address, 0)
```
emersion, line 395: Style: we use tabs.

You can make sure the style is okay by running `go fmt ./...`.

Apart from this LGTM!

kyazdani42: (forgot to run gofmt). I'll try to build hydroxide against this to see if the issue is properly fixed

emersion:
You can run hydroxide with this by using a local copy of go-imap by changing hydroxide's go.mod:

```
replace github.com/emersion/go-imap => /path/to/patched/go-imap
```

kyazdani42:
Ok well the issue is still happening. I need to dig deeper:
as go nil value for arrays is `[](nil)` and not `nil`, when formatting, it returns `[]` instead of `nil`.

emersion:
Maybe making `FormatAddressList` return `interface{}` could help? That way it would return a nil interface, rather than an empty slice.

kyazdani42:
You were right.
I'll update the tests and it should be all good !

write.go
```diff
@@ -195,7 +195,11 @@ func (w *Writer) writeField(field interface{}) error {
 	case Literal:
 		return w.writeLiteral(field)
 	case []interface{}:
-		return w.writeList(field)
+		if len(field) == 0 {
```
emersion, line 198: This is incorrect, an empty slice should be formatted as `()`

kyazdani42: that's what i thought, but it breaks the tests otherwise.

kyazdani42: `expected, _ := formatFields(test.fields)` will format incorrectly

emersion: Not changing something shouldn't break tests that pass.

kyazdani42: except that those tests are failing because now that the envelope can have nil fields, the addresses change so `TestMessage_Format` and `TestEnveloppe_Format` will break.

kyazdani42: Should i create a new writer for the Envelope then ?

kyazdani42: hum my bad, i should just change the test data ...

emersion: Can't we just fix `envelopeTests`? Basically replace `[]interface{}{}` with `nil`.

kyazdani42: should be all good now :)

PR #363, opened 2020-06-19
backendutil: Implement message size and lines counting

foxcpp:


backend/backendutil/bodystructure.go
```diff
@@ -1,14 +1,40 @@
...
 	"mime"
 	"strings"
 
 	"github.com/emersion/go-imap"
 	"github.com/emersion/go-message/textproto"
 )
 
+type countReader struct {
+	r           io.Reader
+	total       uint32
```
emersion, line 17: Maybe consider renaming this to `bytes`, to make it clear what the count is.

backend/backendutil/bodystructure.go
```diff
@@ -1,14 +1,40 @@
...
+	newlines    uint32
+	lastNewline bool
+}
+
+func (r *countReader) Read(b []byte) (int, error) {
+	n, err := r.r.Read(b)
+	r.total += uint32(n)
+	if n != 0 {
+		r.newlines += uint32(bytes.Count(b[:n], []byte{'\r'}))
```
emersion, line 26: Any reason why we count `\r` instead of `\n` here?

backend/backendutil/bodystructure.go
```diff
@@ -48,14 +73,41 @@ func FetchBodyStructure(header textproto.Header, body io.Reader, extended bool)
...
+			bufBody := bufio.NewReader(&countedBody)
+			subMsgHdr, err := textproto.ReadHeader(bufBody)
+			if err != nil {
+				return nil, err
+			}
+			bs.Envelope, err = FetchEnvelope(subMsgHdr)
+			if err != nil {
+				return nil, err
+			}
+			// TODO: Limit recursion.
```
emersion, line 91: I wonder if we should do this in go-message directly. Would probably be a good idea.

backend/backendutil/bodystructure.go
foxcpp: Woops, right. It is supposed to be `\n` but I forgot to change it back after checking something.

backend/backendutil/bodystructure.go
foxcpp: I will drop that TODO, I think we have an issue open in go-message already.

PR #365, opened 2020-07-07
Support NIL heirarchy delimiter

### LINKED ISSUE #329: Library cannot handle NIL hierarchy delimiter
Trying to do `LIST` against an IMAP server that uses a flat structure (NIL hierarchy delimiter) will fail with error `Mailbox delimiter must be a string`.

This is because the parsing of the IMAP response detects a `NIL` atom (at `read.go:220`) and puts a nil interface there in the response fields; this nil interface later fails a type assertion to string (`responses/list.go:35` -> `mailbox.go:63`), and the library rejects this because it enforces that all mailboxes require valid hierarchy delimiters.

However, RFC3501 says a `NIL` hierarchy delimiter is valid and should be used to indicate a flat hierarchy (see [RFC3501#7.2.2](https://tools.ietf.org/html/rfc3501#section-7.2.2))

An example of a significant mail server that reports a `NIL` delimiter is seznam.cz.

A quick fix (though i'm not sure about further sideeffects at this point) is to change `mailbox.go:63` to be like
```go
if info.Delimiter, ok = fields[1].(string); fields[1] != nil && !ok {
    return errors.New("Mailbox delimiter must be a string")
}
```

jameshoulahan:
This fixes #329.

In order to make this a non-breaking change I introduce a boolean to indicate whether the delimiter is NIL or not.

~~Also I hope it's okay to include testify. If not, I'll switch to using standard lib.~~ removed testify

### DIFF
--- a/mailbox.go
+++ b/mailbox.go
@@ -43,7 +43,8 @@ type MailboxInfo struct {
 	// The mailbox attributes.
 	Attributes []string
 	// The server's path separator.
-	Delimiter string
+	Delimiter       string
+	HasNILDelimiter bool
 	// The mailbox name.
 	Name string
 }
@@ -61,7 +62,11 @@ func (info *MailboxInfo) Parse(fields []interface{}) error {
 
 	var ok bool
 	if info.Delimiter, ok = fields[1].(string); !ok {
-		return errors.New("Mailbox delimiter must be a string")
+		// The delimiter may be specified as NIL, which gets converted to a nil interface.
+		if fields[1] != nil {
+			return errors.New("Mailbox delimiter must be a string")
+		}
+		info.HasNILDelimiter = true
 	}
 
 	if name, err := ParseString(fields[2]); err != nil {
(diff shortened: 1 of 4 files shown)


mailbox.go
```diff
@@ -43,7 +43,8 @@ type MailboxInfo struct {
 	// The mailbox attributes.
 	Attributes []string
 	// The server's path separator.
-	Delimiter string
+	Delimiter       string
+	HasNILDelimiter bool
```
emersion, line 47: Instead of adding this new field, is it possible to treat an empty `Delimiter` as NIL? Or would that conflict with something else?

emersion:
>Also I hope it's okay to include testify. If not, I'll switch to using standard lib.

I'd prefer to stick with the stdlib if possible.

jameshoulahan:
> > Also I hope it's okay to include testify. If not, I'll switch to using standard lib.
> 
> I'd prefer to stick with the stdlib if possible.

Agreed, after seeing that this was the only outside dependency, i removed it

mailbox.go
jameshoulahan: Not too sure. I didn't want someone who explicitly wants their delimiter to be `""` (for whatever insane reason) to be surprised when it comes out as `NIL`. I don't think the spec says anything about that.

jameshoulahan: Alright, I think there's no need to worry about a `""` delimiter. This means we can avoid introducing the boolean which is much cleaner imo.

emersion: IMAP says the hierarchy delimiter is a "character", so I think it'd be fine. I'd prefer this to avoid having two sources of truth (what happens if `Delimiter` is non-empty and `HasNILDelimiter` is true?).

PR #366, opened 2020-07-08
Error when selecting should unselect

### LINKED ISSUE #351: Failed select should unselect
From RFC https://tools.ietf.org/html/rfc1730#section-6.3.1:

>      Only one mailbox may be selected at a time in a session;
>      simultaneous access to multiple mailboxes requires multiple
>      sessions.  The SELECT command automatically deselects any
>      currently selected mailbox before attempting the new selection.
>      Consequently, if a mailbox is selected and a SELECT command that
>      fails is attempted, no mailbox is selected.

jameshoulahan:
Fixes #351 

(also includes a little tidyup that the linter spotted)

### DIFF
--- a/server/cmd_auth.go
+++ b/server/cmd_auth.go
@@ -26,15 +26,10 @@ func (cmd *Select) Handle(conn Conn) (err error) {
 	// 		currently selected mailbox before attempting the new selection.
 	// 		Consequently, if a mailbox is selected and a SELECT command that
 	// 		fails is attempted, no mailbox is selected.
-	// Thus if SELECT fails for whatever reason we should unselect.
 	// For example, some clients (e.g. Apple Mail) perform SELECT "" when the
 	// server doesn't announce the UNSELECT capability.
-	defer func() {
-		if err != nil {
-			ctx.Mailbox = nil
-			ctx.MailboxReadOnly = false
-		}
-	}()
+	ctx.Mailbox = nil
+	ctx.MailboxReadOnly = false
 
 	if ctx.User == nil {
 		return ErrNotAuthenticated
--- a/server/cmd_auth.go
+++ b/server/cmd_auth.go
@@ -24,8 +24,21 @@ func (cmd *Select) Handle(conn Conn) error {
 		return ErrNotAuthenticated
 	}
 
+	// As per RFC1730#6.3.1,
+	// 		The SELECT command automatically deselects any
+	// 		currently selected mailbox before attempting the new selection.
+	// 		Consequently, if a mailbox is selected and a SELECT command that
+	// 		fails is attempted, no mailbox is selected.
+	// Thus if this GetMailbox call fails to find a mailbox, this is not an
+	// error and we should simply unselect.
+	// For example, some clients (e.g. Apple Mail) perform SELECT "" when the
+	// server doesn't announce the UNSELECT capability.
 	mbox, err := ctx.User.GetMailbox(cmd.Mailbox)
-	if err != nil {
+	if errors.Is(err, backend.ErrNoSuchMailbox) {
+		ctx.Mailbox = nil
+		ctx.MailboxReadOnly = false
+		return nil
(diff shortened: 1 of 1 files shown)


server/cmd_auth.go
```diff
@@ -24,8 +24,21 @@ func (cmd *Select) Handle(conn Conn) error {
...
+	// 		currently selected mailbox before attempting the new selection.
+	// 		Consequently, if a mailbox is selected and a SELECT command that
+	// 		fails is attempted, no mailbox is selected.
+	// Thus if this GetMailbox call fails to find a mailbox, this is not an
+	// error and we should simply unselect.
+	// For example, some clients (e.g. Apple Mail) perform SELECT "" when the
+	// server doesn't announce the UNSELECT capability.
 	mbox, err := ctx.User.GetMailbox(cmd.Mailbox)
-	if err != nil {
+	if errors.Is(err, backend.ErrNoSuchMailbox) {
```
emersion, line 37: Nice catch! Should we unselect on any error, or just when the mailbox isn't found?

The RFC wording makes me think the former would be better, in which case we don't need to check for `backend.ErrNoSuchMailbox`.

foxcpp: I agree with emersion, there are other reasons SELECT may fail and the effect should be the same per RFC.

jameshoulahan: Good point. I guess it's quite clear -- always unselect first, then only select if no other errors are thrown.

I don't suppose it's threadsafe enough to set Mailbox to nil at the top of the function, which would follow RFC's description even more closely.

jameshoulahan: Hmm but should `Handle` return an error if `GetMailbox` does? Should all errors from `GetMailbox` be ignored? Not sure what's best to do here.

jameshoulahan: Pushed changes that unselect on any SELECT failure; all errors are bubbled up.

emersion: >Hmm but should Handle return an error if GetMailbox does? Should all errors from GetMailbox be ignored? Not sure what's best to do here.

It should be fine to always reset the mailbox, since errors returned here will be forwarded to client via an IMAP response.

server/cmd_auth.go
```diff
@@ -18,15 +18,30 @@ type Select struct {
...
+
+	// As per RFC1730#6.3.1,
+	// 		The SELECT command automatically deselects any
+	// 		currently selected mailbox before attempting the new selection.
+	// 		Consequently, if a mailbox is selected and a SELECT command that
+	// 		fails is attempted, no mailbox is selected.
+	// Thus if SELECT fails for whatever reason we should unselect.
+	// For example, some clients (e.g. Apple Mail) perform SELECT "" when the
+	// server doesn't announce the UNSELECT capability.
+	defer func() {
```
emersion, line 32: This doesn't build.

I think it's fine to just update the context's current mailbox at the top of the function.

emersion: The race you're referring to is a separate bug. Apparently we didn't have an issue for it, filled one: https://github.com/emersion/go-imap/issues/367

jameshoulahan: Done. Now all tests pass.

foxcpp:
Speaking of UNSELECT, perhaps #342 can help. :)

server/cmd_auth.go
```diff
@@ -18,15 +18,25 @@ type Select struct {
 	commands.Select
 }
 
-func (cmd *Select) Handle(conn Conn) error {
+func (cmd *Select) Handle(conn Conn) (err error) {
```
emersion, line 21: One last nit: can you revert the change that switches to named return values?

See this for details: https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

jameshoulahan: oh yes :man_facepalming: sure

PR #374, opened 2020-07-26
go-imap v2 Backend interface work 

### LINKED ISSUE #323: Delegate updates dispatching to the backend

foxcpp:
See messages for individual commits.

Closes #323.


foxcpp:
Okay, describing first commit in a greater detail:

1. backend.Updater is removed and this commit implements #323 with several adjustments. 
  Notably, User interface stays named User. Connection object is passed to GetMailbox instead.

2. backend.Mailbox interface is now solely reserved for operations on selected mailbox. This is 
  important to avoid calling now potentially more expensive GetMailbox for commands like
  SUBSCRIBE, STATUS that could be implemented without the need to establish sequence number
  view and update synchronization dance. This involves moving Status, SetSubscribed methods
  to User interface.

3. Initial Status call for Mailbox is squashed into GetMailbox operation. This is based on the 
  observation that requesting status information may involve operations that are also 
  performed when opening mailbox (establishing sequence numbers view). That is, it is likely 
  inexpensive for some backend implementations to do both operations at the same time.

foxcpp:
Speaking of backendutil helpers, I extracted https://github.com/foxcpp/go-imap-sequpdate from my go-imap-maildir experiments. I am not putting it into this PR as I believe it is not polished out enough yet.

foxcpp:
Note to self: go-imap-namespace needs updating for the new Interface

foxcpp:
Note to self: SendUpdate might need synchronization.

foxcpp:
Another note to self: Need to somehow workaround sending of updates in response to CLOSE. This is potentially harmful deviation from RFC 3501.

foxcpp:
TODO: Documentation should be cleaned from Updater, etc references.

foxcpp:
Just rebased it onto master. If we synchronize dev with master conflicts should be gone.

emersion:

PR #378, opened 2020-08-02
backendutil: Improve Match function

foxcpp:


backend/backendutil/search.go
```diff
@@ -194,11 +216,12 @@ func matchSeqNumAndUid(seqNum uint32, uid uint32, c *imap.SearchCriteria) bool {
 }
 
 func matchDate(date time.Time, c *imap.SearchCriteria) bool {
-	date = date.Round(24 * time.Hour)
+	date = time.Date(date.Year(), date.Month(), date.Day(), 0, 0, 0, 0, time.UTC)
```
emersion, line 219: The date isn't guaranteed to be UTC, is it?

backend/backendutil/search.go
```diff
@@ -66,12 +79,12 @@ func Match(e *message.Entity, seqNum, uid uint32, date time.Time, flags []string
 		if err != nil {
 			return false, err
 		}
-		t = t.Round(24 * time.Hour)
+		t = time.Date(t.Year(), t.Month(), t.Day(), 0, 0, 0, 0, time.UTC)
```
emersion, line 82: Ditto

backend/backendutil/search.go
```diff
@@ -41,16 +42,28 @@ type lengther interface {
...
+	return len(b), nil
+}
+
 func bodyLen(e *message.Entity) (int, error) {
+	headerSize := countWriter{}
+	textproto.WriteHeader(&headerSize, e.Header.Header)
+
 	if l, ok := e.Body.(lengther); ok {
-		return l.Len(), nil
+		return l.Len() + headerSize.N + 2 /* CRLF for empty line */, nil
```
emersion, line 59: Doesn't `WriteHeader` already write the final double CRLF?

backend/backendutil/search.go
```diff
@@ -104,8 +117,17 @@ func Match(e *message.Entity, seqNum, uid uint32, date time.Time, flags []string
...
 	for _, text := range c.Text {
-		// TODO: also match header fields
-		if ok, err := matchBody(e, text); err != nil || !ok {
+		headerMatch := false
+		for f := e.Header.Fields(); f.Next(); {
+			decoded, err := f.Text()
+			if err != nil {
+				continue
+			}
+			if strings.Contains(f.Key()+": "+decoded, text) {
```
emersion, line 126: I don't remember whether this should match raw fields or decoded fields. Which one is it?

backend/backendutil/search.go
foxcpp: Time zone here should be the same as in criteria and actual value is not important - comparison is required to be zone unaware.

backend/backendutil/search.go
foxcpp: Hm, gotcha, you are right.

backend/backendutil/search.go
foxcpp: RFC 3501 says that fields should be decoded if CHARSET is specified. We do not preserve the information about that argument so the only reasonable action is to decode and hope client specified charset (most probably do).

backend/backendutil/search.go
```diff
@@ -194,11 +216,14 @@ func matchSeqNumAndUid(seqNum uint32, uid uint32, c *imap.SearchCriteria) bool {
...
-	date = date.Round(24 * time.Hour)
+	// We discard time zone information by setting it to UTC.
+	// RFC 3501 explicitly requires zone unaware date comparison.
+	date = time.Date(date.Year(), date.Month(), date.Day(), 0, 0, 0, 0, time.UTC)
+
 	if !c.Since.IsZero() && !date.After(c.Since) {
 		return false
 	}
-	if !c.Before.IsZero() && !date.Before(c.Before) {
+	if !c.Before.IsZero() && !date.Before(c.Before) && !date.Equal(c.Before) {
```
emersion, line 226: `Time.Before` already checks that the two dates aren't equal.

backend/backendutil/search.go
emersion: Ah, indeed. Thanks for adding a comment explaining that.

backend/backendutil/search.go
emersion: Makes sense!

backend/backendutil/search.go
```diff
@@ -66,12 +79,12 @@ func Match(e *message.Entity, seqNum, uid uint32, date time.Time, flags []string
...
 			return false, err
 		}
-		t = t.Round(24 * time.Hour)
+		t = time.Date(t.Year(), t.Month(), t.Day(), 0, 0, 0, 0, time.UTC)
 
 		if !c.SentBefore.IsZero() && !t.Before(c.SentBefore) {
 			return false, nil
 		}
-		if !c.SentSince.IsZero() && !t.After(c.SentSince) {
+		if !c.SentSince.IsZero() && !t.After(c.SentSince) && !t.Equal(c.SentSince) {
```
emersion, line 87: Ditto: `Time.After` checks for inequality

foxcpp: Oh, you are right on Before but for After there is more: RFC 3501 names it "since" as it is >= (name in go-imap is somewhat misleading). In this particular case we skip the message if message matches the inverted criteria. But !t.After is <= (inverse of >), while we want inverse of >= which is <, therefore additional condition is needed to make <= into <.

emersion: Ah. Can we use `t.Before` instead of `!t.After` then maybe?

foxcpp: Right, we should (how come i find the least clean solution to the problem every single time?)

Issue #383, opened 2020-08-20
Lowercase DispositionParams keys?


sqwishy:
So I'm using the mail client aerc and sometimes it looks for a "filename" parameter in the content disposition on a message part. In my case, the string keys in the `DispositionParams` in the `imap.Message` are all uppercase. So my email has a "FILENAME" but not a "filename". My guess is that this library is is just using what the imap server gave it in the response?

I don't really know if the mail duder is being a weirdo, or if maybe parsing the BodyStructure should lowercase the keys kind of like how `mime.ParseMediaType` does?

Below is from a debugging session to illustrate the above.

```
> github.com/emersion/go-imap.parseHeaderParamList() /home/sqwishy/go/pkg/mod/github.com/emersion/go-imap@v1.0.4/message.go:123 (PC: 0x972c38)
Values returned:
	~r1: map[string]string [
	        "FILENAME": "Super Important Document.pdf",
	        "SIZE": "1094837",
	        "CREATION-DATE": "Wed, 06 May 2020 19:39:42 GMT",
	        "MODIFICATION-DATE": "Fri, 15 May 2020 22:23:32 GMT",
	]
	~r2: error nil

   118:func encodeHeader(s string) string {
   119:	return mime.QEncoding.Encode("utf-8", s)
   120:}
   121:
   122:func parseHeaderParamList(fields []interface{}) (map[string]string, error) {
=> 123:	params, err := ParseParamList(fields)
   124:	if err != nil {
   125:	        return nil, err
   126:	}
   127:
   128:	for k, v := range params {
(dlv) p fields
[]interface {} len: 8, cap: 8, [
	"FILENAME",
	"Super Important Document.pdf",
	"SIZE",
	"1094837",
	"CREATION-DATE",
	"Wed, 06 May 2020 19:39:42 GMT",
	"MODIFICATION-DATE",
	"Fri, 15 May 2020 22:23:32 GMT",
]
```

foxcpp:
Most strings in MIME and IMAP are case-insensitive. Need to take a look at RFC 3501.

emersion:
>maybe parsing the BodyStructure should lowercase the keys

Yes, I agree.

foxcpp:
> Most strings in MIME and IMAP are case-insensitive. Need to take a look at RFC 3501.

RFC 3501 just refers to RFC 2183, which defines it as case-insensitive via ABNF. But yeah, I agree lower-casing it will make it less surprising. Could you submit a PR to do so?

[...]

PR #384, opened 2020-08-22
imap: lower some fields + content disposition keys

sqwishy:
Regarding #383

`TestBodyStructure_Format` does not pass right now because the BodyStructure contains lowercased values of the input string. So when it's formatted, its field list does not match the input.

I could:
1. Move the case out of `bodyStructureTests` into its own test function that does the equivalent of `TestBodyStructure_Parse` on that one case.
2. Add an optional member to the `bodyStructureTests` element struct type to allow overriding the expected value instead of `formatFields(test.fields)`.

The first option seems less freaky. I thought I'd ask in case you have better ideas or preferences.

### DIFF
--- a/message.go
+++ b/message.go
@@ -119,6 +119,11 @@ func encodeHeader(s string) string {
 	return mime.QEncoding.Encode("utf-8", s)
 }
 
+func stringLowered(i interface{}) (string, bool) {
+	s, ok := i.(string)
+	return strings.ToLower(s), ok
+}
+
 func parseHeaderParamList(fields []interface{}) (map[string]string, error) {
 	params, err := ParseParamList(fields)
 	if err != nil {
@@ -126,8 +131,14 @@ func parseHeaderParamList(fields []interface{}) (map[string]string, error) {
 	}
 
 	for k, v := range params {
+		if lower := strings.ToLower(k); lower != k {
+			delete(params, k)
+			k = lower
+		}
+
 		params[k], _ = decodeHeader(v)
 	}
+
 	return params, nil
 }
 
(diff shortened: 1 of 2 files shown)


message.go
```diff
@@ -119,6 +119,24 @@ func encodeHeader(s string) string {
 	return mime.QEncoding.Encode("utf-8", s)
 }
 
+func toLowerTrimmed(s string) string {
+	return strings.TrimSpace(strings.ToLower(s))
```
emersion, line 123: Do we really need to trim spaces here? I haven't encountered fields with spaces in the wild yet.

sqwishy: Yah I don't know anything about email; wanted to get your guys' thoughts.  I removed it.

message.go
```diff
@@ -911,9 +925,11 @@ func (bs *BodyStructure) Parse(fields []interface{}) error {
 			if disp, ok := fields[end].([]interface{}); ok && len(disp) >= 2 {
 				if s, ok := disp[0].(string); ok {
 					bs.Disposition, _ = decodeHeader(s)
+					bs.Disposition = strings.ToLower(bs.Disposition)
 				}
 				if params, ok := disp[1].([]interface{}); ok {
 					bs.DispositionParams, _ = parseHeaderParamList(params)
+					keysLowered(bs.DispositionParams)
```
emersion, line 932: I think we can inline the lower-casing in `parseHeaderParamList`.

sqwishy: Can we do that in `ParseParamList`? I think that's the most clean as the map is first populated there, but I didn't want to mess with a public/exported function.

emersion: Yeah, we can't do that because that would be a breaking change.

emersion (approved):
LGTM, thanks!

PR #387, opened 2020-09-07
SELECT, EXAMINE and LOGOUT implicitly CLOSE

cuthix:


foxcpp:
Note to self: Make sure this change is properly integrated into go-imap v2 branch.

server/cmd_auth.go
```diff
@@ -26,10 +26,16 @@ func (cmd *Select) Handle(conn Conn) error {
...
 	// 		fails is attempted, no mailbox is selected.
+	//
 	// For example, some clients (e.g. Apple Mail) perform SELECT "" when the
 	// server doesn't announce the UNSELECT capability.
-	ctx.Mailbox = nil
-	ctx.MailboxReadOnly = false
+	//
+	// RFC3501#section-6.4.2 CLOSE
+	// 		The SELECT, EXAMINE, and LOGOUT commands implicitly close the
+	// 		currently selected mailbox without doing an expunge.
```
emersion, line 35: `closeMailbox` calls `Expunge`, but the RFC says we shouldn't.

cuthix: RFC says it should `CLOSE`

The diference between `CLOSE` and `EXPUNGE` is that `EXPUNGE` is verbose while `CLOSE` is not sending any unilateral responses.

The our imap backend interface have just `Expunge`. IMAP server is using this call also for `CLOSE` but it first removes the `context.Mailbox` to not send any updates.

emersion: "without doing an expunge" means without deleting messages with have the `\\Deleted` flag.

cuthix: if this is true then why there is extension `UNSELECT`

https://tools.ietf.org/html/rfc3691

```
2.  UNSELECT Command

   Arguments:  none

   Responses:  no specific responses for this command

   Result:     OK - unselect completed, now in authenticated state
               BAD - no mailbox selected, or argument supplied but
                     none permitted

      The UNSELECT command frees server's resources associated with the
      selected mailbox and returns the server to the authenticated
      state.  This command performs the same actions as CLOSE, except
      that no messages are permanently removed from the currently
      selected mailbox.
```

emersion: Sorry for the delay, I was on leave for a few weeks.

- `CLOSE` should perform an expunge
- `UNSELECT` should *not* perform an expunge
- `SELECT`, `EXAMINE`, and `LOGOUT` should implicitly close the current mailbox *without* performing an expunge (like `UNSELECT`)

The last point doesn't hold with this PR. Does that clarify my comment?

emersion (changes requested):
(See comment above)

emersion:
Ping

emersion:
Fixed in go-imap v2.

PR #390, opened 2020-09-10
fix: import cycle not allowed

proletarius101:
Same as https://github.com/emersion/go-message/pull/109


proletarius101:
Then wait for https://github.com/emersion/go-message/pull/109

emersion:
Does this also fix the same error? Maybe a linter gets executed and makes the packaging fail?

This patch looks good to me if you change the commit message to "utf7: fix package doc comment".

proletarius101:
I guess so (they should really make it more clear). Commit message changed.

emersion:
Thanks!

PR #392, opened 2020-09-14
return empty reader instead of nil when BODY is found but server returns nil

y0ast:
Server can return nil for a body part, when it's empty but still there. RFC3501 says clients should treat this same as empty string. This patch implements that behavior.


emersion (approved):
LGTM, thanks!

PR #394, opened 2020-09-22
add support for empty groups in address lists

y0ast:
Currently when an email contains `To: undisclosed-recipients:;` in the header, `go-imap` parses this as one address with just a mailbox field, and one with all `nill` fields. Calling `.Address()` on them creates an array with `<undisclosed-recipients@>, <@>`.

The reason is that `go-imap` does not support empty groups, see e.g. :

```
Appendix A.1.3.  Group Addresses

   ----
   From: Pete <pete@silly.example>
   To: A Group:Ed Jones <c@a.test>,joe@where.test,John <jdoe@one.test>;
   Cc: Undisclosed recipients:;
   Date: Thu, 13 Feb 1969 23:32:54 -0330
   Message-ID: <testabcd.1234@silly.example>

   Testing.
   ----

   In this message, the "To:" field has a single group recipient named
   "A Group", which contains 3 addresses, and a "Cc:" field with an
   empty group recipient named Undisclosed recipients.
``` 

in https://www.ietf.org/rfc/rfc5322.txt

The proposed solution is to check if an address has a valid mailbox and host name, otherwise return an error and avoid adding the address to the list (instead of ignoring it like now).

Alternatively, the group could be parsed properly, but I have no idea where to start for that (it seems that the server directly returns it in fields, so my solution seems like the first check on content so it's appropriate?).

### DIFF
--- a/message.go
+++ b/message.go
@@ -674,12 +674,18 @@ func (addr *Address) Parse(fields []interface{}) error {
 	if s, err := ParseString(fields[1]); err == nil {
 		addr.AtDomainList, _ = decodeHeader(s)
 	}
-	if s, err := ParseString(fields[2]); err == nil {
-		addr.MailboxName, _ = decodeHeader(s)
+
+	s, err := ParseString(fields[2])
+	if err != nil {
+		return errors.New("Mailbox name could not be parsed")
 	}
-	if s, err := ParseString(fields[3]); err == nil {
-		addr.HostName, _ = decodeHeader(s)
+	addr.MailboxName, _ = decodeHeader(s)
+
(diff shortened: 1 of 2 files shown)


foxcpp (commented):
Could you add a test to verify your example is now handled is a more meaningful way?

y0ast:
No problem, I've just added tests for the two cases.

y0ast:
@emersion I was wondering if you had time to take a look at the above? This problem comes up in `aerc` and deals with a case that's described in the RFC but not handled in go-imap.

foxcpp (approved):
LGTM

PR #398, opened 2020-10-22
client: Drop syscall.ECONNRESET hack from Client.readOnce

### LINKED ISSUE #397: Plan9 support
Apparently syscall.ECONNRESET is undefined when building with GOOS=plan9.
This causes compilation to fail here: https://github.com/emersion/go-imap/blob/c8e5db3cf6c416215b98533fa0996a131574e1dd/client/client.go#L157

Context: We got somebody trying to run [foxcpp/maddy](https://github.com/foxcpp/maddy) on Plan9. Client package is imported indirectly via go-imap-compress:
```
⟩ go mod why github.com/emersion/go-imap/client
# github.com/emersion/go-imap/client
github.com/foxcpp/maddy/internal/endpoint/imap
github.com/emersion/go-imap-compress
github.com/emersion/go-imap/client
```

foxcpp:
Closes #397

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -13,7 +13,6 @@ import (
 	"net"
 	"os"
 	"sync"
-	"syscall"
 	"time"
 
 	"github.com/emersion/go-imap"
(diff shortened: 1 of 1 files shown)


emersion (approved):
Yes, we should find the cause of the ECONNRESET in our tests instead of ignoring it.

emersion:
Thanks!

PR #409, opened 2020-11-30
Fix custom delimiter bug

ben221199:
This Pull Request does fix the bug that makes it impossible to have every custom delimiter.
For example, the character with hex `0x1C` was encoded by UTF-7 and would become `&HA-`.
IMAP clients like Thunderbird then could not detect it being an delimiter.
Therefore, I removed all UTF-7 encoding of the folders and added it to the backend.

Then, if you use the hex `0x1C` as delimiter, then you should encode your folder name in the backend like this:
When you have a folder with name `folder` having a folder `subfolder` having a folder `subsubfolder`, then encode this folder names seperate from each other. Then append this names together with your delimiter (in this case the hex `0x1C`).
You will get `foldersubfoldersubsubfolder`, where `` is your delimiter.

### DIFF
--- a/backend/memory/backend.go
+++ b/backend/memory/backend.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/emersion/go-imap"
 	"github.com/emersion/go-imap/backend"
+	"github.com/emersion/go-imap/utf7"
 )
 
 type Backend struct {
@@ -34,9 +35,11 @@ func New() *Backend {
 		"\r\n" +
 		"Hi there :)"
 
+	inboxName, _ := utf7.Encoding.NewEncoder().String("INBOX")
+
 	user.mailboxes = map[string]*Mailbox{
 		"INBOX": {
-			name: "INBOX",
+			name: inboxName,
 			user: user,
 			Messages: []*Message{
 				{
--- a/commands/append.go
+++ b/commands/append.go
@@ -50,8 +50,8 @@ func (cmd *Append) Parse(fields []interface{}) (err error) {
 	// Parse mailbox name
 	if mailbox, err := imap.ParseString(fields[0]); err != nil {
 		return err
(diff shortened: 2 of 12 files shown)


emersion:
NACK from me. UTF-7 is an IMAP implementation detail and users shouldn't have to deal with it.

ben221199:
TODO:
 - Fix some tests

foxcpp:
NACK from me. It is going to cause further problems if UTF8=ACCEPT/UTF8=ONLY support is added to go-imap.

ben221199:
But you cannot just encode a character if it is used as delimiter...

emersion:
Why not? Where does the RFC says that the delimiter shouldn't be encoded?

foxcpp:
@ben221199 is referring to issues with existing clients. Of course RFC treats the whole mailbox path as a single string so that would not cause issues if clients were not broken.

ben221199:
If I use https://graphemica.com/001C as delimiter and it is encoded by UTF-7. Then it will become `&HA-`.
Then, if the response is `* LIST () "" "folder&HA-subfolder&HA-subsubfolder"`, then Thunderbird will not detect `&HA-` as delimiter.

foxcpp:
Would it work if the delimiter is encoded too?

emersion:
Then report a bug to Thunderbird. In any case, closing this since it's not the direction we want to go in.

foxcpp:
```

mailbox-list    = "(" [mbx-list-flags] ")" SP
                   (DQUOTE QUOTED-CHAR DQUOTE / nil) SP mailbox
QUOTED-CHAR     = <any TEXT-CHAR except quoted-specials> /
                  "\" quoted-specials
quoted-specials = DQUOTE / "\"
TEXT-CHAR       = <any CHAR except CR and LF>
```

ben221199:
By the way, RFC3501 mentions UTF-7 as convention.

PR #417, opened 2021-02-12
utf7: reset ascii boolean after decode while destination is short.

daichitakahashi:
Hello,  
Thanks for this useful library.  
In Japanese environment, I found a bug in decoding UTF-7.

In `decoder.Transform`, there are two cases to return `transform.ErrShortDst`.  
The second case occurs after decode Base64. This is the point.  

When whole decoded string cannot be passed to `dst`, next call of `decoder.Transform` will begin with decoding Base64(repeat last decode and try to copy them to `dst` again).  
So, `decoder.ascii` should be set `true` before return `transform.ErrShortDst`. Otherwise, misdetection of Null shift makes error.

### DIFF
--- a/utf7/decoder.go
+++ b/utf7/decoder.go
@@ -77,9 +77,7 @@ func (d *decoder) Transform(dst, src []byte, atEOF bool) (nDst, nSrc int, err er
 		}
 
 		if nDst+len(b) > len(dst) {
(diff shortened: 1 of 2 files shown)


emersion (approved):
Thanks for the fix!

PR #422, opened 2021-03-27
Fix #421: Correct docstring on client.Append

### LINKED ISSUE #421: Documentation error
The documentation for Client.Append says:

> flags and date are optional arguments and can be set to nil. 

However, `date` is a non-pointer struct type so it cannot be nil.

yujiri8:


emersion:
Thanks

PR #423, opened 2021-03-31
Support for QRESYNC (RFC 7162) only for the client.

MartinNikolovMarinov:
Hello everyone. I know there has been a long discussion about IMAP extension support and in particular for QRESYNC (RFC 7162). From what I understand you have a plan to implement it in v2, but for some of us this feature is required today.

Perhaps a compromise could be reached.

Implementing this extension in the IMAP server does take quite a bit of effort (I suppose) but adding support in the client is rather simple. In fact all it takes is adding 1 new `StatusItem` type and storing it in the `MailboxStatus` struct. These changes would, almost a 100%, be required if/when you decide to implement it in v2.

Please consider merging this PR so that we can avoid forking and diverging source history further. There are 2 teams in my company that have separate quite diverged forks of this project.

note: 
  There is a change in `go.mod`, which is not related/necessary, but it's done to include this memory efficiency change 
  https://github.com/emersion/go-textwrapper/commit/9839c92a89e1a7bd26071bcb068512e780b4680a
  I realize it adds a bunch of code from `go-message` that might not be wanted, so I am ok with reverting this.
  
Thanks in advance and let me know :)


MartinNikolovMarinov:
Hey guys, any chance for this PR to get merged soon ?

emersion:
This implements only a small part of RFC 7162. How is it useful alone?

emersion:
> There is a change in go.mod, which is not related/necessary

Yeah, please don't include unrelated changes in your PR.

PR #429, opened 2021-04-30
Add issue template

emersion:
Reject issues for questions about go-imap and IMAP in general. Our
issue tracker is filled with questions that aren't actionable.

cc @foxcpp

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,5 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Questions
+    url: "https://webchat.freenode.net/##emersion"
+    about: "Please ask questions in ##emersion on Freenode"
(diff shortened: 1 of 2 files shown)


foxcpp:
I would add a webchat link to help people who are not particularly familiar with IRC.

emersion:
The `config.yml` should already display a link -- here's a preview of how that looks in Sway's issue tracker for instance:



> GitHub introduced "discussions" for that but looks like it is not available for repos as it is "beta"

I guess we _could_ enable it, but I wouldn't personally want to participate, watch or moderate it.

foxcpp:
>A link in a text input isn't really clickable, so I haven't included it in the template.

Woops, I am a bit blind.

emersion:
No worries :P alright, let's merge this and see how it goes.

emersion:
Well, not so well… :/



Will push a fixup.

PR #431, opened 2021-05-11
Fetch: Ignore unilateral updates from the server

### LINKED ISSUE #430: FETCH FLAGS and unilateral messages: Unexpected results
IMAP servers (here: Dovecot) tend to send you some information "while you're at it". go-imap in general handles them via the `Updates` channel. I stumbled now upon an issue with FETCH FLAGS and an unilateral update:

```
a66 uid fetch 213531 flags
* 2004 FETCH (UID 213531 FLAGS (\Seen))
* 1994 FETCH (FLAGS ())
a66 OK Fetch completed (0.001 + 0.000 secs).
```

In this case, a `c.UidFetch(213531, imap.FetchFlags)` _(pseudo-code)_ will yield TWO messages as a result (the one for the message asked, and another). And this was totally unexpected on my end, leading to dead-locks, because I stopped reading from the message channel after the first result.

This is either

- an implementation bug: `Fetch.Handle` should ensure to only handle requested messages and pass the rest onto the Update handler
- a documentation ... enhancement: Make it clear, that this command can return more than one asked for.

If it is decided to be the former, I can create a PR for the fix.

Necoro:
The server can send unilateral updates as part of fetch results. Those should be ignored by Fetch and be handled by the `Updates` channel.

This fixes #430.

### DIFF
--- a/responses/fetch.go
+++ b/responses/fetch.go
@@ -34,7 +34,7 @@ func (r *Fetch) Handle(resp imap.Resp) error {
 	}
 
 	if r.Uid && msg.Uid == 0 {
-		// we requested UIDs and got a message without --> unilateral update --> ignore
+		// we requested UIDs and got a message without one --> unilateral update --> ignore
 		return ErrUnhandled
 	}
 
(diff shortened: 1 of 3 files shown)


responses/fetch.go
```diff
@@ -31,6 +33,25 @@ func (r *Fetch) Handle(resp imap.Resp) error {
...
+	var num uint32
+	if r.Uid {
+		num = msg.Uid
+	} else {
+		num = seqNum
+	}
+
+	// check whether we obtained a result we requested with our SeqSet
+	// If it does not, but the msg contains a UID and the set is dynamic (i.e. * or n:*), the server
+	// supplied us with the max UID. That is fine.
```
emersion, line 50: Hm. What do you mean by "the server supplied us with the max UID"?

Necoro: From the RFC, I got that when requesting a dynamic UID sequence like "4:*" the server can answer with the message with UID "2" iff this is the maximum UID. (Reasoning: "4:*" == "*:4")

emersion: Ah. Indeed:

```
                    ; Example: a unique identifier sequence range of
                    ; 3291:* includes the UID of the last message in
                    ; the mailbox, even if that value is less than 3291.
```

Can you expand the comment a bit?

Necoro: I did (see fresh commit): Is this to your liking?

```go
// Check whether we obtained a result we requested with our SeqSet
// If the result is not contained in our SeqSet we have to handle an additional special case:
// In case we requested UIDs with a dynamic sequence (i.e. * or n:*) and the maximum UID of the mailbox
// is less than our n, the server will supply us with the max UID (cf. RFC 3501 §6.4.8 and §9 `seq-range`).
// Thus, such a result is correct and has to be returned by us.
```

emersion (approved):
Thanks!

PR #432, opened 2021-05-11
Update dependencies

Necoro:
Dependencies are bit out of date.

Btw: Deps and required go-versions between the different emersion/* packages are sometimes inconsistent. For instance, go-message (HEAD, not 0.14.1) requires Go version 1.14, while this project remains with 1.13.


foxcpp:
Thanks!

PR #453, opened 2021-12-22
Add IDLE to capabilities list

sblinch:
IDLE was changed into a built-in extension, but the default capabilities list was never updated to include it. The go-imap/client library (and probably any other sane IMAP client) resorts to polling if the IDLE cap isn't present. This PR simply adds IDLE to the list.

### DIFF
--- a/server/conn.go
+++ b/server/conn.go
@@ -163,7 +163,7 @@ func (c *conn) Close() error {
 }
 
 func (c *conn) Capabilities() []string {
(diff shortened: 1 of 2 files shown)


emersion:
`IDLE` is only supported if the server implements it.

emersion:
Hm, scratch that, we should always support it indeed.

Can you update the tests accordingly?

PR #463, opened 2022-04-29
Fix FETCH command with ALL, FAST and FULL macro as fetch items

LBeernaertProton:
When using one of the macros as fetch item arguments, the IMAP command
written to the socket would wrap the macro in parentheses, which is not
valid.

To address this we check whether there is only one item in the fetch
list and if it matches on of the macros we make sure it's serialized in
way that does not wrap it in parentheses.

### DIFF
--- a/commands/fetch.go
+++ b/commands/fetch.go
@@ -14,14 +14,22 @@ type Fetch struct {
 }
 
 func (cmd *Fetch) Command() *imap.Command {
-	items := make([]interface{}, len(cmd.Items))
-	for i, item := range cmd.Items {
-		items[i] = imap.RawString(item)
-	}
+	// Handle FETCH macros separately as they should not be serialized within parentheses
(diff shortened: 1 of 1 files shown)


commands/fetch.go
```diff
@@ -14,14 +14,22 @@ type Fetch struct {
...
 func (cmd *Fetch) Command() *imap.Command {
-	items := make([]interface{}, len(cmd.Items))
-	for i, item := range cmd.Items {
-		items[i] = imap.RawString(item)
-	}
+	// Handle FETCH macros separately as they should not be serialized within parentheses
+	if len(cmd.Items) == 1 && (cmd.Items[0] == imap.FetchAll || cmd.Items[0] == imap.FetchFast || cmd.Items[0] == imap.FetchFull) {
+		return &imap.Command{
+			Name:      "FETCH",
+			Arguments: []interface{}{cmd.SeqSet, cmd.Items},
```
emersion, line 21: Shouldn't this be `imap.RawString(cmd.Items[0])`?

LBeernaertProton: Good catch! Fixed.

emersion (approved):
LGTM, thanks!

PR #465, opened 2022-05-19
search: allow users to select us-ascii charset

konimarti:
Add a UseUSASCII bool flag to the search criteria to indicate that this
charset should be used for the search command. Default charset remains
utf-8.

According to RFC 3501, the search command must only support us-ascii;
support for other charset is optional.

Currently, us-ascii is the fallback charset when the search command with
utf-8 returns a badcharset error. However, this means that we always
execute two search calls to servers that don't support utf-8. More
importantly, it often leads to a server-side disconnect after multiple
badcharset errors. This can be avoid with the proposed flag.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -83,6 +83,12 @@ type Client struct {
 	// access.
 	locker sync.Mutex
 
+	// This flag is set when the first search query fails with a BADCHARSET
+	// error. Subsequent queries will be performed with the US-ASCII
+	// charset. According to RFC 3501, SEARCH must only support US-ASCII;
+	// other charsets are optional.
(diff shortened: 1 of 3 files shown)


emersion:
I'd rather not do this. If the goal is to not send multiple UTF-8 search queries when the server doesn't support them, I'd prefer to store a `utf8SearchUnsupported` flag in `Client`, set when the first search query fails.

konimarti:
Thanks for the feedback @emersion I'll change the code and push a new commit with your proposal.

emersion (approved):
Thanks!

PR #471, opened 2022-08-15
Adding ON for search criteria

alexk307:
Adds support for searching for messages by `ON` as specified in the IMAP spec

```
      ON <date>
         Messages whose internal date (disregarding time and timezone)
         is within the specified date.
```
https://datatracker.ietf.org/doc/html/rfc3501#section-6.4.4

### DIFF
--- a/search.go
+++ b/search.go
@@ -67,6 +67,7 @@ type SearchCriteria struct {
 	Before     time.Time // Internal date is before this date
 	SentSince  time.Time // Date header field is since this date
 	SentBefore time.Time // Date header field is before this date
+	On         time.Time // Date header field is on this date
 
 	Header textproto.MIMEHeader // Each header field value is present
 	Body   []string             // Each string is in the body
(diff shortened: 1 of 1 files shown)


emersion:
What is the motivation for this?

alexk307:
Hey @emersion!  In order to search for messages between two dates, you need to specify both the `SINCE`, and the `BEFORE`. The specification is that `SINCE` is inclusive, and `BEFORE` is exclusive. In the case where you'd want to search for all messages received on a certain day, you need to be aware of the this fact, and set the `BEFORE` parameter to tomorrow's date.

By using the `ON` parameter, you can just specify one date to grab messages from.

alexk307:
Happy to fix up the failing test if this approach is cool with you 👍

alexk307:
any interest @emersion ?

emersion:
I'd rather have a `SetOn` method which sets `SINCE` and `BEFORE` I think.

alexk307:
It's supported by the RFC - so I thought a library that aims to support IMAP would want to implement it 👍 

https://datatracker.ietf.org/doc/html/rfc1730#page-27

PR #477, opened 2022-09-28
Enable remove quote mark

christianWilling:
fixes that the `ENABLE` command adds quote mark to the capabilities


emersion (approved):
Thanks!

PR #488, opened 2023-03-07
v2: add basic client

emersion:
- [ ] Hangs when a parse error happens in a FETCH (`FetchCommand` channel not closed)
- [ ] Recover from panics in decoder goroutine


imapclient/decode.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func readMsgAtt(dec *imapwire.Decoder, cmd *FetchCommand) error {
+	if !dec.ExpectSpecial('(') {
+		return dec.Err()
+	}
+
+	items := make(chan *FetchItem, 32)
+	defer close(items)
+	cmd.msgs <- &FetchData{items: items}
```
emersion: `cmd` can be nil here

imapclient/client.go
```diff
@@ -0,0 +1,564 @@
...
+
+// Wait blocks until the command has completed.
+func (cmd *Command) Wait() error {
+	if cmd.enc != nil {
+		panic("command waited before being closed")
+	}
+	if cmd.encErr != nil {
+		return cmd.encErr
+	}
+	return <-cmd.done
```
emersion: We need to allow `Wait` to be called multiple times

imapclient/client.go
```diff
@@ -0,0 +1,768 @@
...
+	for i := range c.contReqs {
+		if (<-chan struct{})(c.contReqs[i]) == ch {
+			c.contReqs = append(c.contReqs[:i], c.contReqs[i+1:]...)
+			break
+		}
+	}
+	c.mutex.Unlock()
+}
+
+func (c *Client) encodeLiteral(size int64) io.WriteCloser {
```
emersion: If the server doesn't reply with a continuation request, but instead replies with an error, we need to unregister the continuation request.

emersion:
Hm, weird, GitHub auto-closed this PR, and I cannot re-open it now. Continuing in #490.

PR #495, opened 2023-03-24
[v2] Add basic server

emersion:
- [x] Send `OK [CAPABILITY]`, `APPENDUID`, `COPYUID`
- [x] Send continuation requests for literals outside of `APPEND`
- [ ] Plumb context
- [x] Timeouts
- [x] Debug writer
- [x] Retrofit IMAP4rev1 support
- [x] Sort out unilateral updates
- [x] Add support for `FETCH`, `FLAGS`, unilateral updates
- [ ] Add support for `CREATE`, `DELETE`, `RENAME`, , `PERMANENTFLAGS` unilateral updates (maybe more?)
- [x] Add `Server.Close`
- [ ] Add `Server.Shutdown`
- [ ] Add `PREAUTH` support
- [ ] Add support for `SEARCHRES`

### DIFF
--- a/append.go
+++ b/append.go
@@ -1,0 +1,16 @@
+package imap
+
+import (
+	"time"
+)
+
+// AppendOptions contains options for the APPEND command.
+type AppendOptions struct {
+	Flags []Flag
+	Time  time.Time
+}
+
+// AppendData is the data returned by an APPEND command.
+type AppendData struct {
+	UID, UIDValidity uint32 // requires UIDPLUS or IMAP4rev2
+}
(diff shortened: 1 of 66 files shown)


emersion:
Regarding unilateral updates the following decisions have been made, but I'm not sure they're a good idea:

- `Session.Poll` is invoked at the end of every command. The alternative would be to only invoke it on `NOOP` and let the backend manually do the polling in each command's handler.
- It is not possible to write arbitrary updates to a `Conn` at any time. Backends can only send unilateral updates in `Poll` and `Idle`.
- `Session.Expunge` and `Session.Store` are passed in a writer to send responses to the client. However to properly behave when the same mailbox is selected from multiple connections, backends should send the updates in `Poll` instead. The `.SILENT` modifier for `STORE` is supposed to send updates to other clients only and is not handled. Maybe there should be a way to queue a `MailboxTracker` update on all connections except the current one?

PR #496, opened 2023-03-30
fix #412

### LINKED ISSUE #412: aliyun mail uses UidSearch() or Search() method to report "invalid command or parameters"
Hello, why does the program report an error "invalid command or parameters" when I use aliyun mail to call Search() or UidSearch(), below is my code and error info

        var c *client.Client
        // imap.mxhichina.com:993
	server := param.Server.Server 
	email := param.Server.Email
	password := param.Server.Password
	folder := param.Folder

	c = connect(server, email, password)
	if c == nil {
		return nil
	}

	_, err := c.Select(folder, true)

	if err != nil {
		log.Fatal(err)
	}

	// Set search criteria
	criteria := imap.NewSearchCriteria()
	criteria.WithoutFlags = []string{imap.SeenFlag}
	ids, err := c.Search(criteria)
	if err != nil {
		log.Fatal(err)
	}
	log.Println("IDs found:", ids)

 
image

qjebbs:
This PR should fix #412, it helps when the server doesn't allow `CHARSET` in `SEARCH` command and reports `BAD`.

The RFC seems to suggest that `US-ASCII` is the default charset, so the PR omit it to support those not-well-implemented servers. (Yes, you, qiye.aliyun.com)

### DIFF
--- a/client/cmd_selected.go
+++ b/client/cmd_selected.go
@@ -119,12 +119,13 @@ func (c *Client) executeSearch(uid bool, criteria *imap.SearchCriteria, charset 
 func (c *Client) search(uid bool, criteria *imap.SearchCriteria) (ids []uint32, err error) {
 	charset := "UTF-8"
 	if c.utf8SearchUnsupported {
-		charset = "US-ASCII"
+		charset = ""
 	}
 	ids, status, err := c.executeSearch(uid, criteria, charset)
-	if status != nil && status.Code == imap.CodeBadCharset {
-		// Some servers don't support UTF-8
-		ids, _, err = c.executeSearch(uid, criteria, "US-ASCII")
+	if status != nil &&
+		(status.Code == imap.CodeBadCharset || // Some servers don't support UTF-8
+			status.Type == imap.StatusRespBad) { // Some servers don't even support CHARSET within COMMAND
+		ids, _, err = c.executeSearch(uid, criteria, "")
 		c.utf8SearchUnsupported = true
 	}
 	return
--- a/client/cmd_selected_test.go
+++ b/client/cmd_selected_test.go
@@ -178,7 +178,7 @@ func TestClient_Search_Badcharset(t *testing.T) {
 	s.WriteString(tag + " NO [BADCHARSET (US-ASCII)]\r\n")
 
 	// internal fall-back to US-ASCII which sets utf8SearchUnsupported = true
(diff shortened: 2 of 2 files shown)


emersion:
Hm, I don't see anywhere in the RFC that US-ASCII is the default…

qjebbs:
I found these lines in [RFC 3501](https://www.rfc-editor.org/rfc/rfc3501) which says the `CHARSET` is OPTIONAL.

---

[6.4.4](https://www.rfc-editor.org/rfc/rfc3501#section-6.4.4).  SEARCH Command

   Arguments:  OPTIONAL [[CHARSET](https://www.rfc-editor.org/rfc/rfc3501#ref-CHARSET)] specification
               searching criteria (one or more)

   Responses:  REQUIRED untagged response: SEARCH

   Result:     OK - search completed
               NO - search error: can't search that [[CHARSET](https://www.rfc-editor.org/rfc/rfc3501#ref-CHARSET)] or
                    criteria
               BAD - command unknown or arguments invalid

emersion:
Right, but it doesn't say anything about the default… In go-imap v2 I only specify the charset when the search contains UTF-8 FWIW.

qjebbs:
See also:

Characters are 7-bit US-ASCII unless otherwise specified.  Other
   character sets are indicated using a "CHARSET", as described in
   [[MIME-IMT](https://www.rfc-editor.org/rfc/rfc3501#ref-MIME-IMT)] and defined in [[CHARSET](https://www.rfc-editor.org/rfc/rfc3501#ref-CHARSET)].  CHARSETs have important
   additional semantics in addition to defining character set; refer to
   these documents for more detail.

So, if no otherwise specified, it's US-ASCII. That's what `default` exactly means.

emersion:
That's fine if the search criteria only contains ASCII. If it contains characters out of ASCII, then who knows what might happen. (That's why v2 is forcing UTF-8 when the search criteria isn't ASCII.)

qjebbs:
> 



> That's fine if the search criteria only contains ASCII. If it contains characters out of ASCII, then who knows what might happen. (That's why v2 is forcing UTF-8 when the search criteria isn't ASCII.)

See. I don't familiar with all these RFCs at all, is there any thing we can do like what it does to encode email headers?

```
From: =?utf-8?q?Needs_Enc=C3=B3ding?= <encoding@example.com>
```

PR #497, opened 2023-04-03
imapmemserver: send FLAGS updates

emersion:
- [ ] During `STORE`, send new mailbox flags before `FETCH` responses


emersion:
Maybe we could introduce `StoreWriter`, which is `FetchWriter` with an additional function to send new FLAGS. `StoreWriter` is responsible for sending the FLAGS update at the right time (before FETCH).

PR #499, opened 2023-04-04
ci: run imaptest

### LINKED ISSUE #243: Run ImapTest

emersion:
Closes: https://github.com/emersion/go-imap/issues/243

### DIFF
--- a/.build.yml
+++ b/.build.yml
@@ -1,6 +1,10 @@
 image: alpine/latest
 packages:
   - go
+  # for imaptest
+  - curl
+  - gcc
+  - netcat-openbsd
 sources:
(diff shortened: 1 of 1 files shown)


emersion:
Alternatively we should be able to grab an imaptest binary from https://github.com/dovecot/imaptest/releases/tag/latest

foxcpp:
FWIW for maddy-storage, I wrote several helper functions that run imaptest scripted tests in subtests. Makes it convenient to just `go test` the backend without any extra steps.

https://github.com/foxcpp/maddy-storage/blob/dev/tests/imap2/imaptest_test.go

PR #505, opened 2023-04-07
Make SearchCriteria.Not a pointer

emersion:


seqset.go
```diff
@@ -317,3 +317,26 @@ func (s SeqSet) search(q uint32) (i int, ok bool) {
 	}
 	return min, s[min].Contains(q)
 }
+
+func intersectSeqSet(a, b SeqSet) SeqSet {
+	// TODO: optimize
+	var out SeqSet
+	for _, seq := range a {
+		for _, other := range b {
+			// TODO: handle '*'
```
emersion: This is actually tricky to handle. For instance:

```
42 EXISTS
SEARCH 1:* 1:2
-> 1:2
SEARCH 1:* 1:60
-> 1:*
SEARCH * 20
-> nil
SEARCH * 42
-> *
```

I think we need to either accumulate multiple seqsets per criteria, either track the number of messages in the selected mailbox in the server connection logic.

emersion: Removed the need for this in #523

search.go
```diff
@@ -65,7 +65,11 @@ func (criteria *SearchCriteria) And(other *SearchCriteria) {
 		criteria.Smaller = other.Smaller
 	}
 
-	criteria.Not = append(criteria.Not, other.Not...)
+	if criteria.Not != nil && other.Not != nil {
+		criteria.Not.And(other.Not)
```
emersion, line 69: This is incorrect. Should be `Or`.

emersion:
Doesn't seem like this is worth the effort.

PR #508, opened 2023-04-08
feat: support noop command write resp

chyroc:
```
tag1 noop
* 459 EXISTS
* 419 RECENT
tag1 OK NOOP completed
tag2 noop
* 460 EXISTS
* 224 RECENT
tag2 OK NOOP completed
```


emersion:
I'd recommend using v2 if you're looking for a non-broken unilateral update mechanism.

PR #511, opened 2023-04-12
Keep caps in cache to avoid Literal to be stuck

mariwilliam:
When Literal() is called from append command, the server capabilities may not have been retrieved yet. Therefore, a call to Caps() may trigger the encoding of the capability command which will cause a deadlock (we are already encoding the append command). To solve this, we call Caps() before the append command to ensure it is properly cached.

### DIFF
--- a/imapclient/append.go
+++ b/imapclient/append.go
@@ -13,6 +13,8 @@ import (
 //
 // The options are optional.
 func (c *Client) Append(mailbox string, size int64, options *imap.AppendOptions) *AppendCommand {
+	// Keep caps in cache to avoid beginCommand to be stuck
+	c.Caps()
 	cmd := &AppendCommand{}
 	cmd.enc = c.beginCommand("APPEND", cmd)
 	cmd.enc.SP().Mailbox(mailbox).SP()
--- a/imapclient/append.go
+++ b/imapclient/append.go
@@ -13,8 +13,6 @@ import (
 //
 // The options are optional.
 func (c *Client) Append(mailbox string, size int64, options *imap.AppendOptions) *AppendCommand {
-	// Keep caps in cache to avoid beginCommand to be stuck
-	c.Caps()
 	cmd := &AppendCommand{}
 	cmd.enc = c.beginCommand("APPEND", cmd)
 	cmd.enc.SP().Mailbox(mailbox).SP()
--- a/imapclient/client.go
+++ b/imapclient/client.go
@@ -962,7 +962,10 @@ func (ce *commandEncoder) flush() {
 // Literal encodes a literal.
 func (ce *commandEncoder) Literal(size int64) io.WriteCloser {
 	var contReq *imapwire.ContinuationRequest
-	if size > 4096 || !ce.client.caps.Has(imap.CapLiteralMinus) {
+	ce.client.mutex.Lock()
+	hasCapLiteralMinus := ce.client.caps.Has(imap.CapLiteralMinus)
+	ce.client.mutex.Unlock()
+	if size > 4096 || !hasCapLiteralMinus {
 		contReq = ce.client.registerContReq(ce.cmd)
 	}
 	ce.client.setWriteTimeout(literalWriteTimeout)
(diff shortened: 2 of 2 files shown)


emersion:
`beginCommand` specifically does not use `Client.Caps` to avoid this issue, instead it looks up the cap map `c.cap` directly.

Can you elaborate on the issue you're seeing?

mariwilliam:
The following code is still using the client to call Caps() (we are facing deadlocks on mutex at this point):

https://github.com/emersion/go-imap/blob/b1f900fe6ec7c25b13f8dc4e180fdd7accf198ba/imapclient/client.go#L965

emersion:
Oh. Can we change it to lookup the map directly?

mariwilliam:
Done.

imapclient/client.go
```diff
@@ -962,7 +962,7 @@ func (ce *commandEncoder) flush() {
 // Literal encodes a literal.
 func (ce *commandEncoder) Literal(size int64) io.WriteCloser {
 	var contReq *imapwire.ContinuationRequest
-	if size > 4096 || !ce.client.Caps().Has(imap.CapLiteralMinus) {
+	if size > 4096 || !ce.client.caps.Has(imap.CapLiteralMinus) {
```
emersion: We need to acquire the client mutex when looking up `client.caps`.

imapclient/append.go
```diff
@@ -13,6 +13,8 @@ import (
 //
 // The options are optional.
 func (c *Client) Append(mailbox string, size int64, options *imap.AppendOptions) *AppendCommand {
+	// Keep caps in cache to avoid beginCommand to be stuck
+	c.Caps()
```
emersion: Now that we don't use `Client.Caps` anymore, this can be removed.

mariwilliam: The thing is that client.caps may be nil, which is why I added this call at first (this function acquires the mutex properly and populates client.caps if nil).

imapclient/client.go
mariwilliam: I see, therefore calling Caps() may be a better solution.

emersion: No, calling `Caps` may result in a roundtrip, which is undesirable here.

imapclient/append.go
emersion: `Literal` might get called in any other command as well, it's not limited to APPEND.

mariwilliam: Wich solution do you propose to don't be stuck?
The problem is that Append() calls beginCommand() that aqcuires the encoder mutex and then calls Literal() that is stuck because of calling again beginCommand() wich tries to acquire again the encoder mutex.

emersion: I've suggested to stop calling `Caps` in `commandEncoder.Literal`, instead locking `c.mutex` and looking up `c.caps` directly.

mariwilliam: c.caps are nil when calling Literal this why I started Append() with c.Caps().

emersion: It's fine if the caps are nil.

imapclient/client.go
mariwilliam: Mutex acquired in [f5a4d3b](https://github.com/emersion/go-imap/pull/511/commits/f5a4d3bad67798435d9b67e7dea9e56e770e9fdf)

imapclient/append.go
mariwilliam: My server has the "LiteralMinus" capability.
Because caps are nil ce.client.caps.Has(imap.CapLiteralMinus) is false instead of true.
I think caps should be initialized as soon as possible.

emersion: Right, I think we should automatically send a `CAPABILITY` command (but not wait for it to complete) after the greeting/authentication/STARTTLS, if the server didn't send the capabilities on its own.

emersion: I don't think we should wait for capabilities though, as the cost for that is one roundtrip, ie. exactly the same as disabling `LITERAL-`.

emersion (approved):
LGTM, thanks!

imapclient/append.go
emersion: Does https://github.com/emersion/go-imap/pull/514 help with your server?

mariwilliam: Yes, the caps are populated correctly before calling Literal

PR #512, opened 2023-04-12
Add search for Gmail thread Id.

mariwilliam:
Gmail has implemented thread id (https://developers.google.com/gmail/imap/imap-extensions?hl=en#access_to_the_thread_id_x-gm-thrid)
This is a basic way to search messages with the Gmail thread id.


emersion:
I would rather not add support for non-standard Google extensions.

mariwilliam:
Alright, I thought other proprietary imap implementations may use a similar feature which we could "unify" around this parameter at some point.

mariwilliam:
FYI I just added the ability to retrieve thread id from an email, I still believe this can be implemented as is and improved/extended later on with additional imap providers.

emersion:
Sorry, but I think I'll reject this feature for now.

PR #518, opened 2023-04-18
Set IMAP Client timeout options.

mariwilliam:
User wanting to use its own timeout options can set them by populating Options before dialing with IMAP server.
Added PtrT() to get a pointer from any type.

### DIFF
--- a/imapclient/client.go
+++ b/imapclient/client.go
@@ -50,6 +50,12 @@ type Options struct {
 	UnilateralDataHandler *UnilateralDataHandler
 	// Decoder for RFC 2047 words.
 	WordDecoder *mime.WordDecoder
+	// Timeout options
+	IdleReadTimeout     *time.Duration
+	RespReadTimeout     *time.Duration
+	LiteralReadTimeout  *time.Duration
+	CmdWriteTimeout     *time.Duration
+	LiteralWriteTimeout *time.Duration
 }
 
 func (options *Options) wrapReadWriter(rw io.ReadWriter) io.ReadWriter {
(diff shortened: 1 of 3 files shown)


imapclient/client.go
```diff
@@ -1062,3 +1093,8 @@ type loginCommand struct {
 type logoutCommand struct {
 	cmd
 }
+
+// PtrT Get a pointer to any type
+func PtrT[T any](arg T) *T {
```
emersion, line 1098: - I don't want to expose this function in the library's API.
- I don't think the use of generics is warranted here, since the function is only called with one type.
- Just using `&` instead of calling this function works fine.

imapclient/client.go
```diff
@@ -50,6 +50,12 @@ type Options struct {
 	UnilateralDataHandler *UnilateralDataHandler
 	// Decoder for RFC 2047 words.
 	WordDecoder *mime.WordDecoder
+	// Timeout options
+	IdleReadTimeout     *time.Duration
+	RespReadTimeout     *time.Duration
+	LiteralReadTimeout  *time.Duration
+	CmdWriteTimeout     *time.Duration
+	LiteralWriteTimeout *time.Duration
```
emersion, line 58: I'd prefer to not use pointers here, and have:

- Zero means default timeout
- Negative value means no timeout

imapclient/client.go
```diff
@@ -84,6 +90,28 @@ func (options *Options) unilateralDataHandler() *UnilateralDataHandler {
 	return options.UnilateralDataHandler
 }
 
+func initTimeOut(options *Options) *Options {
```
emersion, line 93: Nit: no uppercase of "o" in "timeout"

imapclient/client.go
```diff
@@ -158,6 +186,8 @@ func DialTLS(address string, options *Options) (*Client, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	options = initTimeOut(options)
```
emersion, line 190: This should be performed in `New`.

emersion:
I'm unsure we want to stick with these different kinds of timeouts.

emersion:
Closing for now, please re-open if you want to resume work on this.

PR #522, opened 2023-04-30
imapserver: convert to static SeqSet in core

emersion:
This will help with #505.

- [ ] Can core keep up with the max UID?


emersion:
This seems a bit fragile, maybe better to go with #523 instead.

PR #531, opened 2023-06-29
imapclient: add support for MULTIAPPEND

emersion:


emersion:
Maybe we want to unify the regular APPEND and the MULTIAPPEND codepaths, this would make CATENATE easier to support…

PR #532, opened 2023-07-02
imapclient: fix FetchItems to end encoded list

adregner:
#517 uses the new `enc.BeginList()` to encode fetch options over the IMAP protocol, but never calls `.End()` so the closing `)` character is never written out, causing errors like `T6 BAD Could not parse command`.


emersion (approved):
Eh, I wonder how I managed to not catch this... Thanks for the fix!

PR #538, opened 2023-08-07
imapclient: add SPECIAL-USE LIST option support

rockorager:
Servers are allowed to optionally send SPECIAL-USE mailbox attributes. Add support for clients to explicitly request that SPECIAL-USE attributes be sent.

### DIFF
--- a/list.go
+++ b/list.go
@@ -5,10 +5,12 @@ type ListOptions struct {
 	SelectSubscribed     bool
 	SelectRemote         bool
 	SelectRecursiveMatch bool // requires SelectSubscribed to be set
+	SelectSpecialUse     bool // requires SPECIAL-USE
 
 	ReturnSubscribed bool
 	ReturnChildren   bool
 	ReturnStatus     *StatusOptions // requires IMAP4rev2 or LIST-STATUS
+	ReturnSpecialUse bool           // requires SPECIAL-USE
 }
 
 // ListData is the mailbox data returned by a LIST command.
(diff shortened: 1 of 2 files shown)


list.go
```diff
@@ -5,10 +5,12 @@ type ListOptions struct {
 	SelectSubscribed     bool
 	SelectRemote         bool
 	SelectRecursiveMatch bool // requires SelectSubscribed to be set
+	SelectSpecialUse     bool
```
emersion: Can we add a `// requires SPECIAL-USE` comment here to indicate that bare IMAP4rev1 servers don't support this? (Same for `ReturnSpecialUse`)

rockorager: I put `// requires IMAP4rev2 or SPECIAL-USE`, since SPECIAL-USE was folded into rfc9051

emersion: I don't believe IMAP4rev2 supports these LIST options. The grammar doesn't include them, and appendix E mentions "the list of new mailbox attributes from SPECIAL-USE" (as opposed to just the SPECIAL-USE ext).

rockorager: The [section](https://www.rfc-editor.org/rfc/rfc9051.html#name-list-response) for list response describes the attributes. Both the LIST command and response sections don't at all say you must use SPECIAL-USE to obtain the attributes, but the example shown in the [NAMESPACE section](https://www.rfc-editor.org/rfc/rfc9051.html#section-6.3.10-26) *does* explicitly use a `SPECIAL-USE` selector.

emersion: Yeah, I've noticed. I believe this is copy-pasta from the NAMESPACE ext, I've reported an errata: https://www.rfc-editor.org/errata/eid7593

rockorager: ack. Updating comments accordingly

emersion (approved):
Thanks!

PR #543, opened 2023-08-29
imapclient: add IDLE fallback

emersion:
Not sure we really want this TBH. Are there any servers out there that don't support IDLE?


emersion:
Let's not do this for now. Can always be added later on if need be.

PR #554, opened 2023-10-30
imapserver: split command handling

emersion:
Split into two parts: parsing and processing. Some commands are incompatible with this model (e.g. APPEND), so will have a no-op second step.

The goal is to parallelize the second step with other processing: reading the next command, and eventually handling the next command as well.


emersion:
The main use-case for command pipelining is avoiding to stall the connection when executing long-running commands (e.g. a large SEARCH command). For FETCH the client can split up commands into bite-sized pieces.

Maybe we should only implement this for SEARCH?

PR #555, opened 2023-11-03
v1: run and fix linters

Rikkuru:


emersion:
I don't think most of these changes are desirable.

Previous discussion: https://github.com/emersion/go-imap/pull/370

PR #560, opened 2023-11-22
add rfc 8457 mailbox attr

Rikkuru:
I noticed that "\Important" attr is missing but keyword "$Important" is present
https://www.rfc-editor.org/rfc/rfc8457.html

### DIFF
--- a/imap.go
+++ b/imap.go
@@ -60,13 +60,14 @@ const (
 	MailboxAttrRemote        MailboxAttr = "\\Remote"
 
 	// Role (aka. "special-use") attributes
-	MailboxAttrAll     MailboxAttr = "\\All"
-	MailboxAttrArchive MailboxAttr = "\\Archive"
-	MailboxAttrDrafts  MailboxAttr = "\\Drafts"
-	MailboxAttrFlagged MailboxAttr = "\\Flagged"
-	MailboxAttrJunk    MailboxAttr = "\\Junk"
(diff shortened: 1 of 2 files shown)


imap.go
```diff
@@ -60,13 +60,14 @@ const (
...
-	MailboxAttrSent    MailboxAttr = "\\Sent"
-	MailboxAttrTrash   MailboxAttr = "\\Trash"
+	MailboxAttrAll       MailboxAttr = "\\All"
+	MailboxAttrArchive   MailboxAttr = "\\Archive"
+	MailboxAttrDrafts    MailboxAttr = "\\Drafts"
+	MailboxAttrFlagged   MailboxAttr = "\\Flagged"
+	MailboxAttrJunk      MailboxAttr = "\\Junk"
+	MailboxAttrSent      MailboxAttr = "\\Sent"
+	MailboxAttrTrash     MailboxAttr = "\\Trash"
+	MailboxAttrImportant MailboxAttr = "\\Important" // RFC 8457
```
emersion, line 70: `\Important` is not a role ("special-use") attribute, so can we add a blank line in-between `MailboxAttrTrash` and `MailboxAttrImportant` to make this clear?

emersion: Hm, actually, it is… Nevrmind!

emersion (approved):
Thank you!

PR #563, opened 2023-12-06
imapclient: Close returned err.NetClosed after Logout

Rikkuru:
server closes connection after Logout and Close should not return error NetClosed when called after logout

### DIFF
--- a/imapclient/client.go
+++ b/imapclient/client.go
@@ -22,6 +22,7 @@ package imapclient
 import (
 	"bufio"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"io"
 	"mime"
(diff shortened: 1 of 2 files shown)


emersion:
Can we avoid using `%w` everywhere? That makes the underlying errors part of our API.

Rikkuru:
> part

yes

emersion:
Hm, it sounds like the test succeeds without the bugfix…

Rikkuru:
> Hm, it sounds like the test succeeds without the bugfix…

That's strange. I was able to reproduce it locally 
I created a PR with test without this fix , lets see if test fail there too 
https://builds.sr.ht/~emersion/job/1110697

Rikkuru:
https://github.com/emersion/go-imap/pull/567#issuecomment-1847277690

emersion:
Thanks!

The test is still unreliable, but better than nothing.

PR #569, opened 2023-12-12
imapclient: test unread body

Rikkuru:
Test that shows that client will get stuck on close if body was not read (Close uses Next)

### DIFF
--- a/imapclient/client_test.go
+++ b/imapclient/client_test.go
@@ -3,6 +3,7 @@ package imapclient_test
 import (
 	"io"
 	"net"
+	"strings"
 	"testing"
 
 	"github.com/emersion/go-imap/v2"
@@ -16,11 +17,24 @@ const (
 	testPassword = "test-password"
 )
 
+const simpleRawMessage = `MIME-Version: 1.0
+Message-Id: <191101702316132@example.com>
+Content-Transfer-Encoding: 8bit
+Content-Type: text/plain; charset=utf-8
+
+This is my letter!`
+
 func newClientServerPair(t *testing.T, initialState imap.ConnState) (*imapclient.Client, io.Closer) {
 	memServer := imapmemserver.New()
 
 	user := imapmemserver.NewUser(testUsername, testPassword)
 	user.Create("INBOX", nil)
+
+	_, err := user.Append("INBOX", strings.NewReader(simpleRawMessage), &imap.AppendOptions{})
+	if err != nil {
+		t.Fatalf("Append() = %v", err)
+	}
+
 	memServer.AddUser(user)
 
 	server := imapserver.New(&imapserver.Options{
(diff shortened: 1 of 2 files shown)


imapclient/client_test.go
```diff
@@ -16,11 +18,34 @@ const (
...
+	return fInfo.Size()
+}
+
 func newClientServerPair(t *testing.T, initialState imap.ConnState) (*imapclient.Client, io.Closer) {
 	memServer := imapmemserver.New()
 
 	user := imapmemserver.NewUser(testUsername, testPassword)
 	user.Create("INBOX", nil)
+
+	emlFile, err := os.OpenFile("../testdata/sample.eml", os.O_RDONLY, 0)
```
emersion: Can we maybe use `strings.NewReader` with a multi-line string literal? That should reduce the boilerplate a bit. Also the test message doesn't need to be very complicated, just one or two header and a 1-line body should be enough.

emersion:
The logic to handle this is supposed to be https://github.com/emersion/go-imap/blob/fddf2bba124d44c7efbc906d12b2be85cb65cbde/imapclient/fetch.go#L176

Rikkuru:
> go-imap/imapclient/fetch.go

when is cmd.prev assigned ?

Rikkuru:
Did you have this in mind ?

imapclient/client_test.go
```diff
@@ -14,13 +15,27 @@ import (
 const (
 	testUsername = "test-user"
 	testPassword = "test-password"
+
+	simpleEml = `MIME-Version: 1.0
+	Message-Id: <191101702316132@example.com>
```
emersion: This leaves a tab before the message header fields, which parsers will choke on.

imapclient/client_test.go
```diff
@@ -14,13 +15,27 @@ import (
...
+	Content-Transfer-Encoding: 8bit
+	Content-Type: text/lain; charset=utf-8
```
emersion: Typo: "plain"

imapclient/client_test.go
```diff
@@ -14,13 +15,27 @@ import (
...
 func newClientServerPair(t *testing.T, initialState imap.ConnState) (*imapclient.Client, io.Closer) {
 	memServer := imapmemserver.New()
 
 	user := imapmemserver.NewUser(testUsername, testPassword)
 	user.Create("INBOX", nil)
+
+	emlReader := strings.NewReader(simpleEml)
+	_, err := user.Append("INBOX", emlReader, &imap.AppendOptions{})
+	if err != nil {
+		t.Fatalf("Append simpleEml to INBOX = %s", err)
```
emersion: `%s` is incorrect here, because the value printed has the type `error`.

emersion:
> Did you have this in mind ?

Yes, good find!

I've pushed a new version which fixes my review comments.

Thanks for the fix and test case!

imapclient/client_test.go
Rikkuru: actually it is ok to print errors with %s
s looks if value has interface Stringer or Error
https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/fmt/print.go;l=658

So this 2 ways of printing are equal
```
t.Fatalf("Append simpleEml to INBOX = %s", err.Error())
```
```
t.Fatalf("Append simpleEml to INBOX = %s", err)
```

%v does the same , but %s seems like better choice to me since we expect to see a string returned by Error() method, maybe I should switch to using first example with %s and explicit Error call

emersion: Hm right. But in general I prefer to just use `%v`. `%s` doesn't have any benefit, we don't care about the type here.

PR #575, opened 2023-12-23
imapclient: Allow empty Content-Transfer-Encoding in text parts

### LINKED ISSUE #545: v1: empty encoding in BodyStructure gets serialized as NIL
i assume the encoding should never be specified as an empty string; per RFC 2045:
```
     encoding := "Content-Transfer-Encoding" ":" mechanism

     mechanism := "7bit" / "8bit" / "binary" /
                  "quoted-printable" / "base64" /
                  ietf-token / x-token
```

nonetheless, the library accepts this scenario and the resulting `BODYSTRUCTURE` list cannot be parsed back:
```
* 11319 FETCH (UID 31137 BODYSTRUCTURE ("text" "plain" () NIL NIL NIL 6002 79 NIL NIL NIL NIL) ENVELOPE ("Sun, 10 Sep 2023 02:00:00 +0000" "acme-update" ((NIL NIL "notify" "ptrc.gay")) ((NIL NIL "notify" "ptrc.gay")) ((NIL NIL "notify" "ptrc.gay")) ((NIL NIL "ptrc" "ptrc.gay")) NIL NIL NIL NIL) FLAGS (\Seen nonjunk \Recent))
in response-data: in body-type-1part: imapwire: expected string, got "N"
```

rakoo:
Some messages have some parts with no `Content-Transfer-Encoding`. It's probably illegal, but I have some real world email with that, so ...

2 questions:
- do you plan to have some corpus of emails to double-check ? A lot might be broken, but we still need this library to be useful, so that might solve unforeseen issues
- github is kinda crap, I wish `go-imap` moved to sourcehut :)

Fixes #545

### DIFF
--- a/imapclient/client.go
+++ b/imapclient/client.go
@@ -323,7 +323,7 @@ func (c *Client) Close() error {
 	}
 
 	<-c.decCh
-	if err := c.decErr; err != nil && !errors.Is(err, net.ErrClosed) {
+	if err := c.decErr; err != nil {
 		return err
 	}
 
@@ -519,7 +519,8 @@ func (c *Client) read() {
 
 	c.setReadTimeout(idleReadTimeout)
 	for {
-		if c.dec.EOF() {
+		// Ignore net.ErrClosed here, because we also call conn.Close in c.Close
+		if c.dec.EOF() || errors.Is(c.dec.Err(), net.ErrClosed) {
 			break
 		}
 		if err := c.readResponse(); err != nil {


imapclient/fetch.go
```diff
@@ -804,9 +804,16 @@ func readBodyType1part(dec *imapwire.Decoder, typ string, options *Options) (*im
...
 
 	var description string
-	if !dec.ExpectSP() || !dec.ExpectNString(&bs.ID) || !dec.ExpectSP() || !dec.ExpectNString(&description) || !dec.ExpectSP() || !dec.ExpectString(&bs.Encoding) || !dec.ExpectSP() || !dec.ExpectNumber(&bs.Size) {
+	if !dec.ExpectSP() || !dec.ExpectNString(&bs.ID) || !dec.ExpectSP() || !dec.ExpectNString(&description) || !dec.ExpectSP() || !dec.ExpectNString(&bs.Encoding) || !dec.ExpectSP() || !dec.ExpectNumber(&bs.Size) {
 		return nil, dec.Err()
 	}
+
+	// Content-Transfer-Encoding should always be set, but sometimes it isn't.
+	// Default to 7bit
+	if bs.Encoding == "" && bs.Type == "text" {
```
emersion: The encoding defaults to "7bit" even if the type isn't text.

emersion:
Just checked, and indeed the IMAP grammar doesn't allow NIL here.

emersion:
Does this only happen with the go-imap server? If so, I'd rather fix it there than fix the client.

rakoo:
Actually it happens on the commercial server I use for my personal account (migadu). I'd need to check with their POP interface to see if the email misses `Content-Transfer-Encoding` or if it's a problem with their IMAP issues

I guess some IMAP servers are being lenient and allow this to happen because access to content is more important ? I'll do some investigation

emersion:
Okay. In any case I'd be fine with merging a workaround.

imapclient/fetch.go
rakoo: Done

rakoo:
Just a slight correction: migadu actually does the right thing here, when there is no `Content-Transfer-Encoding` the bodystructure still has a `7bit`. It's another commercial server that does this, but I'd prefer not naming it.

rakoo:
And it looks like you already made the workaround in the server: https://github.com/emersion/go-imap/commit/508db109#diff-59c1c3380d7ee13473234cc454e0128cbbb147fe31740477c4ee8b158b4844faR486

emersion:
Hm, sorry, it seems like I messed up a force-push to this branch to slightly edit the comment. Anyways, I've merged this manually.

Thanks for the fix!

PR #584, opened 2024-01-14
Find filename in upper or lower case and decode it

rakoo:
This PR fixes two issues:

- sometimes in the bodystructure the key is in upper case; the lib couldn't find it
- sometimes the filename is written as RFC2047-encoded words

### DIFF
--- a/imapclient/fetch.go
+++ b/imapclient/fetch.go
@@ -803,15 +803,10 @@ func readBodyType1part(dec *imapwire.Decoder, typ string, options *Options) (*im
 	if !dec.ExpectSP() || !dec.ExpectString(&bs.Subtype) || !dec.ExpectSP() {
 		return nil, dec.Err()
 	}
-
 	var err error
-	bs.Params, err = readBodyFldParam(dec)
+	bs.Params, err = readBodyFldParam(dec, options)
 	if err != nil {
 		return nil, err
-	}
-	if name, ok := bs.Params["name"]; ok {
-		// TODO: handle error
-		bs.Params["name"], _ = options.decodeText(name)
 	}
 
 	var description string
@@ -950,7 +945,7 @@ func readBodyExtMpart(dec *imapwire.Decoder, options *Options) (*imap.BodyStruct
 	var ext imap.BodyStructureMultiPartExt
 
 	var err error
-	ext.Params, err = readBodyFldParam(dec)
+	ext.Params, err = readBodyFldParam(dec, options)
 	if err != nil {
 		return nil, fmt.Errorf("in body-fld-param: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


emersion:
Thanks for the patch!

> sometimes in the bodystructure the key is in upper case; the lib couldn't find it

Maybe we should lower-case keys when populating the map?

> sometimes the filename is written as RFC2047-encoded words

Ideally we'd be using `Client.options.decodeText` so that users can add support for additional charsets (by default `mime.WordDecoder` only supports UTF-8).

rakoo:
Simon Ser ***@***.***> wrote:
> Maybe we should lower-case keys when populating the map?

Good idea, this'll make it available for other fields as well

> > sometimes the filename is written as RFC2047-encoded words
> 
> Ideally we'd be using `Client.options.decodeText` so that users can add support for additional charsets (by default `mime.WordDecoder` only supports UTF-8).

Ah indeed ! I'll use that instead


-- 
Matthieu Rakotojaona

imapclient/fetch.go
```diff
@@ -997,23 +1004,24 @@ func readBodyFldDsp(dec *imapwire.Decoder, options *Options) (*imap.BodyStructur
...
-	}
-
 	if !dec.ExpectSpecial(')') {
 		return nil, dec.Err()
 	}
 	return &disp, nil
 }
 
-func readBodyFldParam(dec *imapwire.Decoder) (map[string]string, error) {
+func readBodyFldParam(dec *imapwire.Decoder, mimeDec func(input string) string) (map[string]string, error) {
```
emersion: Can we maybe just pass the options directly instead of a function?

rakoo:
I didn't know what would be possible, and I was kinda afraid passing the options directly would mess up the clean separation of functionalities. But if that's ok for you, I'll do it :)

-- 
Matthieu Rakotojaona

emersion:
Yeah, I'd prefer to pass the options. It's an internal function anyways.

rakoo:
Updated

imapclient/fetch.go
```diff
@@ -1030,7 +1020,11 @@ func readBodyFldParam(dec *imapwire.Decoder) (map[string]string, error) {
 			if params == nil {
 				params = make(map[string]string)
 			}
-			params[k] = s
+			decoded, err := options.DecodeText(s)
+			if err != nil {
```
emersion: Can we ignore the error for now, and keep the `// TODO: handle error`?

The reason is that decoding errors shouldn't be fatal, but we have no good way to bubble them up currently.

imapclient/fetch.go
```diff
@@ -1030,7 +1020,11 @@ func readBodyFldParam(dec *imapwire.Decoder) (map[string]string, error) {
 			if params == nil {
 				params = make(map[string]string)
 			}
-			params[k] = s
+			decoded, err := options.DecodeText(s)
```
emersion: CI is unhappy, seems like there is a typo here, should be `decodeText`.

rakoo: Oops yes made a typo, sorry

emersion:
Note that it's not strictly correct to decode all parameters, as encoded words are only allowed for some parameters and should be left as-is for others… But we do the same in go-message for simplicity's sake and haven't received a bug report about this yet.

emersion (approved):
Thanks!

rakoo:
Thanks for your patience !

Are there really cases where some values are encoded but we don't want them decoded ? I can't imagine any, but let's see when that happens

PR #585, opened 2024-01-15
imapclient: deadlock on error before greeting

Rikkuru:
I was testing timeouts and got a deadlock in WaitGreeting

### DIFF
--- a/imapclient/client_test.go
+++ b/imapclient/client_test.go
@@ -142,3 +142,19 @@ func TestFetch_closeUnreadBody(t *testing.T) {
 		t.Fatalf("UIDFetch().Close() = %v", err)
 	}
 }
+
+func TestWaitGreeting_eof(t *testing.T) {
+	// bad server: connected but without greeting
+	clientConn, serverConn := net.Pipe()
+
+	client := imapclient.New(clientConn, nil)
+	defer client.Close()
+
+	if err := serverConn.Close(); err != nil {
+		t.Fatalf("serverConn.Close() = %v", err)
+	}
+
+	if err := client.WaitGreeting(); err == nil {
+		t.Fatalf("WaitGreeting() should fail")
+	}
+}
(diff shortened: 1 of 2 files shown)


imapclient/client_test.go
```diff
@@ -62,6 +63,27 @@ func newClientServerPair(t *testing.T, initialState imap.ConnState) (*imapclient
 	return client, server
 }
 
+// bad server: connected but without greeting
+func newClientWithBadServer(t *testing.T) (*imapclient.Client, io.Closer) {
+	ln, err := net.Listen("tcp", "localhost:10004")
```
emersion: Can we use the 0 port, to pick an unused port automatically?

imapclient/client_test.go
```diff
@@ -62,6 +63,27 @@ func newClientServerPair(t *testing.T, initialState imap.ConnState) (*imapclient
...
+
+	conn, err := net.Dial("tcp", ln.Addr().String())
+	if err != nil {
+		t.Fatalf("net.Dial() = %v", err)
+	}
+
+	client := imapclient.New(conn, nil)
+
+	go func() {
+		time.AfterFunc(time.Millisecond, func() { ln.Close() })
```
emersion: No need to wait here, we can close the listener right away since the client already established a connection off of it.

TBH maybe we should be using `net.Pipe` instead?

imapclient/client_test.go
```diff
@@ -62,6 +63,27 @@ func newClientServerPair(t *testing.T, initialState imap.ConnState) (*imapclient
 	return client, server
 }
 
+// bad server: connected but without greeting
+func newClientWithBadServer(t *testing.T) (*imapclient.Client, io.Closer) {
```
emersion: Since this is only used by a single example, and unlikely to be used by any other, maybe we should fold this function's logic into `TestWaitGreetingError`?

emersion:
Overall the bugfix looks good, good catch!

Rikkuru:
@emersion  I changed fix a little 
decErr is not always set 
idk if I should include it at all in WaitGreeting ?

emersion:
Thanks!

PR #589, opened 2024-02-01
fetch: ensure field names are Atoms

rakoo:
This PR makes sure the fields in FETCH/SEARCH are atoms.

The spec says it is allowed to have strings (ie with double quotes) here, but I'm on an IMAP server that doesn't totally respect that:

```
> A004 SEARCH HEADER Message-Id <ea-mime-REDACTED@REDACTED.localdomain>
< * SEARCH 1
< A004 OK SEARCH done
```

```
> A004 FETCH 1 (BODY[HEADER.FIELDS (Message-ID)])
< * 1 FETCH (BODY[HEADER.FIELDS (MESSAGE-ID)] {65}
< Message-ID: <ea-mime-REDACTED@REDACTED.localdomain>
< 
< )
< A004 OK FETCH done
```

```
> A004 FETCH 1 (BODY[HEADER.FIELDS ("Message-ID")])
< * 1 FETCH (BODY[HEADER.FIELDS ("MESSAGE-ID")] {2}
< 
< )
< A004 OK FETCH done
```

The server in question is provided by mailo.com. I know this deviates from standard, but I can only guess if we send something stricter than expected then it's ok ?

### DIFF
--- a/imapclient/fetch.go
+++ b/imapclient/fetch.go
@@ -101,7 +101,8 @@ func writeFetchItemBodySection(enc *imapwire.Encoder, item *imap.FetchItemBodySe
 
 		if len(headerList) > 0 {
 			enc.SP().List(len(headerList), func(i int) {
-				enc.String(headerList[i])
+				// This should be a string, but some servers don't recognize that
+				enc.Atom(headerList[i])
 			})
 		}
 	}


emersion:
Unfortunately this approach breaks the case where values contain invalid atom characters, such as spaces for instance.

Oh well. I suppose we could have `Encoder.AString` which uses an atom if the value doesn't contain special characters?

rakoo:
Wait, there are header field names with spaces ? This change is only for field names, not values (which should always be a string as you say)

-- 
Matthieu Rakotojaona

emersion:
> Wait, there are header field names with spaces ?

Technically yes. The only character a header field can't contain is a colon. See `ftext` in [RFC 5322 section 3.6.8](https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.8).

Regardless, the string comes from the user of the library, which means that it can potentially be anything. We shouldn't produce invalid IMAP syntax even if the user provides an invalid header field name.

rakoo:
Out of curiosity, a literal also doesn't work with mailo

-- 
Matthieu Rakotojaona

PR #591, opened 2024-02-15
Added MYRIGHTS and SETACL support

opsxolc:


emersion:
This is part of the [IMAP4 ACL extension](https://www.rfc-editor.org/rfc/rfc2086).

imapclient/myrights.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"fmt"
+
+	"github.com/emersion/go-imap/v2"
+	"github.com/emersion/go-imap/v2/internal/imapwire"
+)
+
+// Myrights sends a MYRIGHTS command.
+func (c *Client) Myrights(mailbox string) *MyrightsCommand {
```
emersion: Nit: I'd find the capitalization `MyRights` to be more readable.

myrights.go
```diff
@@ -0,0 +1,7 @@
+package imap
+
+// MyrightsData is the data returned by the MYRIGHTS command.
+type MyrightsData struct {
+	Mailbox string
+	Rights  string
```
emersion: Can we use a better type instead of `string`?

Maybe something among the lines of:

```go
type RightSet string

type Right byte

const (
    RightLookup = Right('l')
    …
)
```

imapclient/myrights.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func (cmd *MyrightsCommand) Wait() (*imap.MyrightsData, error) {
+	return &cmd.data, cmd.cmd.Wait()
+}
+
+func readMyrights(dec *imapwire.Decoder) (*imap.MyrightsData, error) {
+	var data imap.MyrightsData
+
+	if !dec.ExpectMailbox(&data.Mailbox) || !dec.ExpectSP() || !dec.ExpectText(&data.Rights) {
```
emersion: The ABNF reads:

```abnf
rights          ::= astring
```

So I think we should use `ExpectAString` instead of `ExpectText`?

imapclient/myrights_test.go
```diff
@@ -0,0 +1,147 @@
...
+	serverErrChan := make(chan error)
+	testcaseChan := make(chan testcaseType)
+
+	defer close(testcaseChan)
+
+	// for client
+	clientErrChan := make(chan error)
+
+	// server
+	go func() {
```
emersion: Hmmm… I'm not a fan of the ad-hoc server code, it would be nicer to use our own server impl from `imapserver` instead…

imapclient/myrights.go
opsxolc: fixed

myrights.go
opsxolc: fixed

imapclient/myrights.go
opsxolc: fixed

imapclient/myrights_test.go
opsxolc: I've added support on the server side for MYRIGHTS and SETACL. But right now SETACL is not checking real meaning of rights on a mailbox, just setting, what's told. I hope its fine(

emersion: In https://github.com/emersion/go-imap/pull/592, I've added Dovecot tests. This allows us to have client tests when the go-imap server implementation is missing, for instance:

```go
if _, ok := server.(*imapserver.Server); ok {
	t.Skip("imapserver doesn't support ACL")
}
```

emersion: Also note that the ACL extension isn't exposed by default by Dovecot, so some slight configuration changes would be required to use that for ACL testing: https://doc.dovecot.org/configuration_manual/acl/#acl

imapclient/client.go
```diff
@@ -942,6 +942,11 @@ func (c *Client) readResponseData(typ string) error {
 			return c.dec.Err()
 		}
 		return c.handleQuotaRoot()
+	case "MYRIGHTS":
+		if !c.dec.ExpectSP() {
+			return c.dec.Err()
+		}
+		return c.handleMyrights()
```
emersion: Nit: the capitalization is inconsistent here, should be `handleMyRights`

imapclient/myrights.go
```diff
@@ -0,0 +1,57 @@
+package imapclient
```
emersion: Can we move everything related to the ACL extension into a single `acl.go` file? I don't think the split is worth it here. Other extensions like NAMESPACE are already in a single file even if they define multiple commands.

myrights.go
```diff
@@ -0,0 +1,7 @@
+package imap
```
emersion: Ditto: I think a single `acl.go` file for all ACL-related things would be enough here

imapclient/myrights.go
```diff
@@ -0,0 +1,57 @@
...
+// MyRights sends a MYRIGHTS command.
+func (c *Client) MyRights(mailbox string) *MyRightsCommand {
+	cmd := &MyRightsCommand{mailbox: mailbox}
+	enc := c.beginCommand("MYRIGHTS", cmd)
+	enc.SP().Mailbox(mailbox)
+	enc.end()
+	return cmd
+}
+
+func (c *Client) handleMyrights() error {
```
emersion: Ditto: `handleMyRights`

imapserver/myrights.go
```diff
@@ -0,0 +1,27 @@
+package imapserver
```
emersion: Ditto: I think a single `acl.go` file for all ACL-related things would be enough here

rights.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+type RightsIdentifier string
+
+const RightsIdentifierAnyone = RightsIdentifier("anyone")
+
+// NewRightSet converts rights string into RightSet with validation
+func NewRightSet(rights string) (RightSet, error) {
+	for _, r := range rights {
+		if !strings.ContainsRune(string(AllRights), r) {
```
emersion: The RFC reads:

```
   Rights is a string listing a (possibly empty) set of alphanumeric
   characters, each character listing a set of operations which is being
   controlled. Letters are reserved for ``standard'' rights, listed
   below.  The set of standard rights may only be extended by a
   standards-track document.  Digits are reserved for implementation or
   site defined rights.
```

Thus, I don't think we should reject rights we don't know about. Maybe a new RFC has defined a new right and go-imap hasn't been updated yet, maybe the server is using an implementation-defined right, either way we shouldn't error out.

rights.go
```diff
@@ -0,0 +1,61 @@
...
+			return "", fmt.Errorf("unsupported right: '%v'", string(r))
+		}
+	}
+
+	return RightSet(rights), nil
+}
+
+func (r RightSet) Add(rights RightSet) RightSet {
```
emersion: Note for a potential future patch: as things stand, "lrs" != "srl" even if these two values represent the same right set. Would be nice to have a canonicalization to avoid this footgun.

imapserver/session.go
```diff
@@ -84,6 +84,15 @@ type SessionNamespace interface {
 	Namespace() (*imap.NamespaceData, error)
 }
 
+// SessionACL is an IMAP session which supports ACL extension methods.
+type SessionACL interface {
+	Session
+
+	// TODO: support the rest ACL extension methods
+	SetACL(mailbox string, identifier imap.RightsIdentifier, rights string) error
```
emersion: `rights` can be a `imap.RightSet` here probably?

imapclient/setacl.go
```diff
@@ -0,0 +1,21 @@
+package imapclient
+
+import "github.com/emersion/go-imap/v2"
+
+// SetACL sends a SETACL command.
+func (c *Client) SetACL(mailbox string, identifier imap.RightsIdentifier, rights imap.RightSet) *SetACLCommand {
+	cmd := &SetACLCommand{}
+	enc := c.beginCommand("SETACL", cmd)
+	enc.SP().Mailbox(mailbox).SP().String(string(identifier)).SP().String(string(rights))
```
emersion: Hm, it's a bit weird to pass an invalid `RightSet` here with a "+" or "-" prefix. Maybe we can have something like this instead?

```go
type RightModification byte

const (
	RightModificationReplace = RightModification(0)
	RightModificationAdd = RightModification('+')
	RightModificationRemove = RightModification('-')
)

func (*Client) SetACL(mailbox string, ri RightsIdentifier, rm RightModification, rs RightSet) *SetACLCommand
```

Does this make sense, do you have a different opinion?

imapclient/acl_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+// TestACL runs tests on SetACL and MyRights commands (for now).
+func TestACL(t *testing.T) {
+	client, server := newClientServerPair(t, imap.ConnStateAuthenticated)
+
+	defer client.Close()
+	defer server.Close()
+
+	client.Create("MyFolder", nil)
+	client.Create("MyFolder/Child", nil)
```
emersion: We should handle errors here, or else this could result in some annoying-to-debug situations in tests.

imapclient/acl_test.go
```diff
@@ -0,0 +1,100 @@
...
+		{
+			Mailbox:   "MyFolder/Child",
+			Rights:    "",
+			SetRights: "",
+		},
+	} {
+		// execute SETACL command
+		err := client.SetACL(testcase.Mailbox, testUsername, testcase.SetRights).Wait()
+		if err != nil {
+			t.Fatalf("[%v] SetACL().Wait() error: %v", i+1, err)
```
emersion: Instead of these `[%v]` prefixes, it would be clearer to use named sub-tests with `t.Run`.

imapserver/imapmemserver/mailbox.go
```diff
@@ -32,9 +33,28 @@ func NewMailbox(name string, uidValidity uint32) *Mailbox {
...
+	}
+
+	return rightsCopy
+}
+
+func (mbox *Mailbox) SetRights(identifier imap.RightsIdentifier, rights imap.RightSet) {
+	mbox.mutex.Lock()
+	mbox.rights[identifier] = rights
+	mbox.mutex.Unlock()
+}
```
emersion: Can we not export these methods in our public API for now? I'm not super confident this is what we'll want to expose long-term.

emersion (commented):
Thanks for working on this, overall this looks pretty good to me!

A few comments, some very minor and a few more API design questions.

This needs a rebase, and will probably fail in CI since we're testing against Dovecot now too (you can probably just skip the tests for Dovecot by checking for the  `*dovecotServer` type if you don't want to deal with this stuff).

imapclient/setacl.go
opsxolc: yeah, you are right, thanks

imapclient/acl_test.go
opsxolc: I wasn't aware of it, thanks

rights.go
emersion: Any thoughts about this?

emersion:
It seems like a `vendor/` dir was committed?

emersion:
Also I'd prefer to avoid the `testify` dependency if possble…

acl.go
```diff
@@ -0,0 +1,110 @@
...
+type RightModification byte
+
+const (
+	RightModificationReplace = RightModification(0)
+	RightModificationAdd     = RightModification('+')
+	RightModificationRemove  = RightModification('-')
+)
+
+// NewRights converts rights string into RightModification and RightSet with validation
+func NewRights(rights string, ignoreUnsupported bool) (RightModification, RightSet, error) {
```
emersion: Do we really need `ignoreUnsupported`?

It's set to `false` in the client code, but I think clients should ignore rights they don't support.

rights.go
emersion: Ah, seems like the latest version has something to accommodate for this.

acl.go
```diff
@@ -0,0 +1,110 @@
...
+			newRights += RightSet(right)
+		}
+	}
+
+	return newRights
+}
+
+func (r RightSet) Equal(rights RightSet) bool {
+	for _, right := range r {
+		if !strings.ContainsRune(string(rights), right) {
```
emersion: This is not enough AFAIU: let's say `r = "lr"` and `rights = "lrs"`, the sets are not equal but this function would still return true.

Related: https://github.com/emersion/go-imap/pull/591#discussion_r1497734574

acl.go
```diff
@@ -0,0 +1,110 @@
...
+			newRights += RightSet(r)
+		} else if !ignoreUnsupported {
+			return rm, "", fmt.Errorf("unsupported right: '%v'", r)
+		}
+	}
+
+	return rm, RightSet(newRights), nil
+}
+
+func IsSupportedRight(r rune) bool {
```
emersion: Maybe `IsStandardRight` would be a better name?

What is the use-case for this? Is it for server implementations? If so, maybe `IsStandardRightSet` taking a `RightSet` would be easier to use?

But I wonder if it's useful to expose this, server implementations will need to go through each right they receive anyways so it should be trivial for them to reject rights they don't support…

imapclient/dovecot_test.go
```diff
@@ -17,13 +17,23 @@ func newDovecotClientServerPair(t *testing.T) (net.Conn, io.Closer) {
...
 `
+
 	if err := os.WriteFile(cfgFilename, []byte(cfg), 0666); err != nil {
 		t.Fatalf("failed to write Dovecot config: %v", err)
 	}
 
 	clientConn, serverConn := net.Pipe()
 
+	// this works for me, while doveadm for some reason don't
+	// cmd := exec.Command("/usr/libexec/dovecot/imap", "-c", cfgFilename)
```
emersion: Weird. What is the issue you're running into exactly?

In any case, this commented code shouldn't be committed.

imapclient/status.go
```diff
@@ -9,6 +9,10 @@ import (
 )
 
 func statusItems(options *imap.StatusOptions) []string {
+	if options == nil {
+		options = &imap.StatusOptions{}
+	}
```
emersion: Is this a nil pointer deref you're running into?

If so, can you submit it as a separate PR?

imapclient/dovecot_test.go
opsxolc: I know that it shouldn't be committed, I wanted to ask maybe you could help? Is it working in your environment?(

imapclient/status.go
opsxolc: OK, I got a panic here

emersion:
Sorry, I keep getting more comments for new functionality...

I would suggest trying to upstream your patches step by step if you want me to merge stuff more quickly. That is, stop adding more features and functions, focus on addressing the review comments. But if you are fine with getting continuous feedback for new ACL functions, that's fine by me as well. I don't want you to be frustrated because of the feedback, is all.

I can also help addressing the review comments if you grow tired of it.

imapclient/dovecot_test.go
emersion: Yeah, it's working fine for me. The CI is using Alpine Linux, and my local setup is Arch Linux.

What is the error you're running into? What is your local OS?

acl.go
opsxolc: I want to be sure I got you right, should we return only standard rights here and ignore nonstandard?

acl.go
opsxolc: I wanted to use this to filter nonstandard rights in NewRights function. You are right, this shouldn't be exposed.

acl.go
emersion: I think we should just pass through everything without filtering. Do you see an issue with this, or do you think it would be better to do it differently?

imapclient/status.go
emersion: I've pushed this change as e767b0db12f2.

opsxolc:
Hey! I'm drowning in work right now, but I'll be back with fixes as soon as I can. I really want to contribute to this project)

emersion:
No worries, take your time :)

imapclient/dovecot_test.go
opsxolc: I'm using CentOS 7. The problem is that doveadm for some reason does not show the ACL extension, when direct run of imap binary with the same config works. 

```
doveadm -D -c test.conf exec imap
Debug: Loading modules from directory: /usr/lib64/dovecot
Debug: Module loaded: /usr/lib64/dovecot/lib01_acl_plugin.so
Debug: Loading modules from directory: /usr/lib64/dovecot/doveadm
Debug: Module loaded: /usr/lib64/dovecot/doveadm/lib10_doveadm_acl_plugin.so
Debug: Skipping module doveadm_expire_plugin, because dlopen() failed: /usr/lib64/dovecot/doveadm/lib10_doveadm_expire_plugin.so: undefined symbol: expire_set_deinit (this is usually intentional, so just ignore this message)
Debug: Skipping module doveadm_quota_plugin, because dlopen() failed: /usr/lib64/dovecot/doveadm/lib10_doveadm_quota_plugin.so: undefined symbol: quota_user_module (this is usually intentional, so just ignore this message)
Debug: Skipping module doveadm_fts_lucene_plugin, because dlopen() failed: /usr/lib64/dovecot/doveadm/lib20_doveadm_fts_lucene_plugin.so: undefined symbol: lucene_index_iter_deinit (this is usually intentional, so just ignore this message)
Debug: Skipping module doveadm_fts_plugin, because dlopen() failed: /usr/lib64/dovecot/doveadm/lib20_doveadm_fts_plugin.so: undefined symbol: fts_user_get_language_list (this is usually intentional, so just ignore this message)
Debug: Skipping module doveadm_mail_crypt_plugin, because dlopen() failed: /usr/lib64/dovecot/doveadm/libdoveadm_mail_crypt_plugin.so: undefined symbol: mail_crypt_box_get_pvt_digests (this is usually intentional, so just ignore this message)
* PREAUTH [CAPABILITY IMAP4rev1 LITERAL+ SASL-IR LOGIN-REFERRALS ID ENABLE IDLE SORT SORT=DISPLAY THREAD=REFERENCES THREAD=REFS THREAD=ORDEREDSUBJECT MULTIAPPEND URL-PARTIAL CATENATE UNSELECT CHILDREN NAMESPACE UIDPLUS LIST-EXTENDED I18NLEVEL=1 CONDSTORE QRESYNC ESEARCH ESORT SEARCHRES WITHIN CONTEXT=SEARCH LIST-STATUS BINARY MOVE SNIPPET=FUZZY] Logged in as k.roshchina
```

```
/usr/libexec/dovecot/imap -c test.conf
* PREAUTH [CAPABILITY IMAP4rev1 LITERAL+ SASL-IR LOGIN-REFERRALS ID ENABLE IDLE SORT SORT=DISPLAY THREAD=REFERENCES THREAD=REFS THREAD=ORDEREDSUBJECT MULTIAPPEND URL-PARTIAL CATENATE UNSELECT CHILDREN NAMESPACE UIDPLUS LIST-EXTENDED I18NLEVEL=1 CONDSTORE QRESYNC ESEARCH ESORT SEARCHRES WITHIN CONTEXT=SEARCH LIST-STATUS BINARY MOVE SNIPPET=FUZZY ACL RIGHTS=texk] Logged in as k.roshchina
```

opsxolc: dovecot version is 2.2.36 (1f10bfa63)

opsxolc:
So builds.sr.ht failed due to the issue with ACL extension support as I mentioned in a thread above

imapclient/dovecot_test.go
opsxolc: So... I've removed dovecot that I installed with yum and builded it from source https://github.com/dovecot/core. And it works now somehow. And also it has more caps by default then dovecot that I installed previously. Tests also succeed now in my environment.

```
doveadm -c test.conf exec imap
Error: conn unix:/usr/local/var/run/dovecot/stats-writer: net_connect_unix(/usr/local/var/run/dovecot/stats-writer) failed: Permission denied
imap(k.roshchina): Error: conn unix:/usr/local/var/run/dovecot/stats-writer: net_connect_unix(/usr/local/var/run/dovecot/stats-writer) failed: Permission denied
* PREAUTH [CAPABILITY IMAP4rev1 SASL-IR LOGIN-REFERRALS ID ENABLE IDLE SORT SORT=DISPLAY THREAD=REFERENCES THREAD=REFS THREAD=ORDEREDSUBJECT MULTIAPPEND URL-PARTIAL CATENATE UNSELECT CHILDREN NAMESPACE UIDPLUS LIST-EXTENDED I18NLEVEL=1 CONDSTORE QRESYNC ESEARCH ESORT SEARCHRES WITHIN CONTEXT=SEARCH LIST-STATUS BINARY MOVE SNIPPET=FUZZY PREVIEW=FUZZY PREVIEW STATUS=SIZE SAVEDATE COMPRESS=DEFLATE INPROGRESS LITERAL+ NOTIFY ACL RIGHTS=texk] Logged in as k.roshchina
```

opsxolc: Maybe it's possible to update CI environment as such, so the tests won't fail?

emersion:
The error message from CI is:

```
    acl_test.go:74: create MyFolder.Child error: imap: NO [CANNOT] Character not allowed in mailbox name: '.' (0.001 + 0.000 secs).
```

By default Dovecot uses `.` for mailbox delimiters. a3d367456797a4cf1a9c853ba0e7d75dff67f069 changes it to `/` (to align it with imapmemserver).

acl.go
```diff
@@ -0,0 +1,104 @@
...
+type MyRightsData struct {
+	Mailbox string
+	Rights  RightSet
+}
+
+// GetACLData is the data returned by the GETACL command.
+type GetACLData struct {
+	Mailbox string
+	Rights  map[RightsIdentifier]RightSet
+}
```
emersion: These are only used by the client, so should be in the client package.

acl.go
```diff
@@ -0,0 +1,104 @@
+package imap
+
+import (
+	"strings"
+)
+
+// IMAP4 ACL extension (RFC 2086)
+
+type RightSet string
```
emersion: It's a bit too tempting for users to try to compare `RightSet` with `==` (which compiles because they're strings, but doesn't do the right thing). Maybe we should turn this into a `[]Right`?

acl.go
```diff
@@ -0,0 +1,104 @@
...
+	RightCreate     = Right('c') // CREATE new sub-mailboxes in any implementation-defined hierarchy
+	RightDelete     = Right('d') // STORE DELETED flag, perform EXPUNGE
+	RightAdminister = Right('a') // perform SETACL
+
+	AllStandardRights = RightSet("lrswipcda")
+)
+
+type RightsIdentifier string
+
+const RightsIdentifierAnyone = RightsIdentifier("anyone")
```
emersion: Maybe we could have a function to create a rights identifier from a username? A simple conversion isn't safe since some identifiers are reserved ("anyone" and strings starting with "-").

acl.go
```diff
@@ -0,0 +1,104 @@
...
+type RightModification byte
+
+const (
+	RightModificationReplace = RightModification(0)
+	RightModificationAdd     = RightModification('+')
+	RightModificationRemove  = RightModification('-')
+)
+
+// NewRights converts rights string into RightModification and RightSet with validation
+func NewRights(rights string) (RightModification, RightSet) {
```
emersion: This doesn't need to be part of our public API I think (could be moved to our `internal` package).

emersion (commented):
I've extracted the client parts in https://github.com/emersion/go-imap/pull/618, reworked them per the review comments, and split them into logical commits. Let me know if these changes look good to you!

emersion:
The client parts have been merged in #618, still missing the server part from this PR.

PR #593, opened 2024-02-21
imapclient: add Dial() for unencrypted connections

bitfehler:
I mean, I certainly don't _insist_ that this method should exist, but [something](https://lists.sr.ht/~sircmpwn/tokidoki-devel/patches/49675#%3C20240219172110.381134-1-contact@emersion.fr%3E-47) gave me the impression you might think so :wink:

### DIFF
--- a/imapclient/client.go
+++ b/imapclient/client.go
@@ -209,6 +209,15 @@ func NewStartTLS(conn net.Conn, options *Options) (*Client, error) {
 	return client, nil
 }
 
+// DialInsecure connects to an IMAP server without any encryption at all.
+func DialInsecure(address string, options *Options) (*Client, error) {
+	conn, err := net.Dial("tcp", address)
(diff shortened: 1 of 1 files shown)


emersion:
Ah, right! I was holding off on this one because I prefer unencrypted connections to be explicit, but maybe I'm worrying too much and should just go with the flow of the rest of the Go ecosystem…

bitfehler:
As I said, it's up to you. I can make the comment more explicit, or change the name to `DialInsecure()` or whatnot, or we can close this and I modify your tokidoki-patch to use `New()` directly.

emersion:
Alright, sounds reasonable to me to rename to `DialInsecure`!

bitfehler:
Exquisite choice! :clinking_glasses: :wink:

emersion (approved):
Thanks!

Issue #571, opened 2023-12-14
v2: in capability-data: imapwire: expected atom, got \"\\r\""


uuthink:
Sometimes while fetching a mail with UIDFetch I get the following error:
```
Expected nil, but got: &errors.errorString{s:"in response-data: in capability-data: imapwire: expected atom, got \"\\r\""}
```

Here are the logs
```
* CAPABILITY IMAP4 IMAP4rev1 XLIST MOVE IDLE XAPPLEPUSHSERVICE NAMESPACE CHILDREN ID UIDPLUS 
```

I noticed that there is a space before the ending newline character

I try to modify the SP method to solve this problem. The modification is as follows. Is it correct?
```
func (dec *Decoder) SP() bool {
	if dec.acceptByte(' ') {
		if dec.CRLF() {
			dec.mustUnreadByte()
			return false
		}
		return true
	}

	// Special case: SP is optional if the next field is a parenthesized list
	b, ok := dec.readByte()
	if !ok {
		return false
	}
	dec.mustUnreadByte()
	return b == '('
}
```

emersion:
Seems similar to https://github.com/emersion/go-imap/issues/540

I'm worried about calling `mustUnreadByte` after `CRLF` though. I think I'd prefer peeking at the next byte and checking whether it's `\r` or `\n`.

uuthink:
Thank you very much, I will using `dec.r.Peek(1)` to get the next byte and checking. 
Is this a bug? or is it caused by not strictly following the RFC? and it will fix in the furture?

emersion:
This is a server bug: they are not strictly following the RFC's grammar.

duesee:
Hey @uuthink, can you say which server generated the CAPABILITY response with the trailing space?

emersion:
For reference, cc @duesee 

imap.exmail.qq.com sends:

    * CAPABILITY IMAP4<SP>IMAP4rev1<SP><CRLF>

smtp-n.global-mail.cn sends:

    * SEARCH<SP><CRLF>

emersion:
Fixed in https://github.com/emersion/go-imap/pull/596

Issue #540, opened 2023-08-18
in response: imapwire: expected CRLF, got " "


k773:
Code to reproduce:
```golang
func TestOutlook(t *testing.T) {
	client, e := imapclient.DialTLS("outlook.office365.com:993", &imapclient.Options{DebugWriter: os.Stderr})
	if e != nil {
		panic(e)
	}
	if e = client.Login(username, password).Wait(); e != nil {
		panic(e)
	}
	if _, e = client.Select("INBOX", nil).Wait(); e != nil {
		panic(e)
	}
	if _, e = client.Status("INBOX", &imap.StatusOptions{NumMessages: true}).Wait(); e != nil {
		panic(e)  // <- panics here
	}
}
```
This code produces the following error:
```
panic: in response: imapwire: expected CRLF, got " " [recovered]
	panic: in response: imapwire: expected CRLF, got " "
```

From the debug output I can see that there is a space symbol after the ')' (second line), which go-imap does not expect:
```
T5 STATUS INBOX (MESSAGES)
* STATUS Inbox (MESSAGES 1) 
T5 OK STATUS completed.
```

emersion:
Does https://github.com/emersion/go-imap/pull/541 help?

k773:
Yes, it does. Thank you.

PR #594, opened 2024-02-23
compatible with non-compliant responses with excessive SPACE

qjebbs:
This PR improves compatibility with non-compliant servers, by skipping excessive SPACE between elements and the trailing SPACE

- imap.exmail.qq.com: `* CAPABILITY IMAP4<SP>IMAP4rev1<SP><CRLF>`
- smtp-n.global-mail.cn: `* SEARCH<SP><CRLF>`

related: #571 #540

### DIFF
--- a/imapclient/capability.go
+++ b/imapclient/capability.go
@@ -40,6 +40,11 @@ func (cmd *CapabilityCommand) Wait() (imap.CapSet, error) {
 func readCapabilities(dec *imapwire.Decoder) (imap.CapSet, error) {
 	caps := make(imap.CapSet)
 	for dec.SP() {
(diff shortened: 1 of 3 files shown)


qjebbs:
the second commit `skip all potential excessive SPACEs` tries to solve all similar problems as it suggests, but I am not sure if I do it in the right way.

qjebbs:
It works good in my env, the only question that comes to mind is whether `SP()`,  `ExpectSP()` needs to be renamed (e.g.: `LSP()` for leading Space) for method semantics reasons.

qjebbs:
#596

PR #595, opened 2024-02-27
POC: rewrite imapnum with generic

qjebbs:
This is a POC to rewrite the `imapnum` package with generics.

Benifits:

1. no need to covert from/to numset with "unsafe" package

	```go
	(*imapnum.Set)(unsafe.Pointer(s))
	```
1. no need to code methods for each type if we don't override them
	```go
	// all methods are implemented in the generic type
	type SeqSet = imapnum.Set[uint32]
	```
1. no need to covert results of `Nums()` when override
	original:
    ```go
    func (s UIDSet) Nums() ([]UID, bool) {
		nums, ok := s.numSet().Nums()
		return uidListFromNumList(nums), ok
	}
	```
	with generic:
    ```go
	func (s UIDSet) Nums() ([]UID, bool) {
		return imapnum.Set[UID](s).Nums()
	}
	```

### DIFF
--- a/internal/imapnum/numset.go
+++ b/internal/imapnum/numset.go
@@ -11,14 +11,14 @@ import (
 // represented by setting Start = Stop. Zero is used to represent "*", which is
 // safe because seq-number uses nz-number rule. The order of values is always
 // Start <= Stop, except when representing "n:*", where Start = n and Stop = 0.
-type Range struct {
-	Start, Stop uint32
+type Range[T ~uint32] struct {
+	Start, Stop T
 }
 
 // Contains returns true if the seq-number q is contained in range value s.
 // The dynamic value "*" contains only other "*" values, the dynamic range "n:*"
 // contains "*" and all numbers >= n.
-func (s Range) Contains(q uint32) bool {
+func (s Range[T]) Contains(q T) bool {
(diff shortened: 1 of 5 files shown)


emersion:
This approach doesn't seem great for end-user documentation.

qjebbs:
> This approach doesn't seem great for end-user documentation.

But I don't see any breaking changes to end users. It changes only the internal unexported underlying package imapnum.

emersion:
On GoDoc, users will no longer see the various methods for `UIDSet` and `SeqSet` listed. They will just see the type alias, and will need to dig into the unexported package to find out what methods they can use and how they can iterate on the set.

qjebbs:
Good point. We can implement SeqSet just like how UIDSet currently does instead of an alias.

emersion:
That sounds better. The remaining downside is that the type definitions are more obscure than before now:

```go
// before
type UIDSet []UIDRange

// after
type UIDSet numset.Set[UID]
```

So some internal types are leaked and it's less obvious that library users can `for _, r := range set`.

qjebbs:
Yes, it's kind of give and take. You are the one who ultimately makes the decision.

PR #596, opened 2024-02-28
compatible with non-compliant responses with excessive SPACE

qjebbs:
This PR improves compatibility with non-compliant servers, by skipping excessive SPACE between elements and the trailing SPACE

- imap.exmail.qq.com: `* CAPABILITY IMAP4<SP>IMAP4rev1<SP><CRLF>`
- smtp-n.global-mail.cn: `* SEARCH<SP><CRLF>`

It works good in my env, the only question that comes to mind is whether `SP()`,  `ExpectSP()` needs to be renamed (e.g.: `LSP()` for leading Space) for method semantics reasons.

related: #571 #540

### DIFF
--- a/internal/imapwire/decoder.go
+++ b/internal/imapwire/decoder.go
@@ -135,7 +135,13 @@ func (dec *Decoder) Expect(ok bool, name string) bool {
 
 func (dec *Decoder) SP() bool {
 	if dec.acceptByte(' ') {
-		return true
+		// https://github.com/emersion/go-imap/issues/571
+		b, ok := dec.readByte()
+		if !ok {
+			return false
+		}
+		dec.mustUnreadByte()
+		return b != '\r' && b != '\n'
 	}
 
 	// Special case: SP is optional if the next field is a parenthesized list


emersion:
Hm, I'd prefer to only diverge from the standard where it's actually needed by some servers. For instance if you haven't hit servers sending `<SP><SP>` I'd prefer to keep rejecting that. Otherwise we might be seeing more servers breaking the standard in more new ways…

emersion:
> It works good in my env, the only question that comes to mind is whether SP(), ExpectSP() needs to be renamed (e.g.: LSP() for leading Space) for method semantics reasons.

I'd prefer not to reflect non-standard server quirks in the API.

internal/imapwire/decoder.go
```diff
@@ -133,9 +133,13 @@ func (dec *Decoder) Expect(ok bool, name string) bool {
...
 }
 
+// SP consumes all following spaces and returns true if there followed by more data.
+// Note, it may still consume spaces (trailing ones) even if it returns false.
 func (dec *Decoder) SP() bool {
 	if dec.acceptByte(' ') {
-		return true
+		for dec.acceptByte(' ') {
+		}
+		return !dec.endOfLine()
```
emersion: Can we add a `// https://github.com/emersion/go-imap/issues/571` comment to explain why this is done here?

internal/imapwire/decoder.go
```diff
@@ -144,15 +148,27 @@ func (dec *Decoder) SP() bool {
 		return false
 	}
 	dec.mustUnreadByte()
+	// sure not end of line
 	return b == '('
 }
 
+// endOfLine returns true if the next bytes are a CR, LF or EOF.
```
emersion: The comment seems out of sync with the implementation here? AFAIU false is returned on EOF?

qjebbs:
> Hm, I'd prefer to only diverge from the standard where it's actually needed by some servers. For instance if you haven't hit servers sending `<SP><SP>` I'd prefer to keep rejecting that. Otherwise we might be seeing more servers breaking the standard in more new ways…

Updated. 

Your opinion makes sense from a library maintainer's perspective, but from users' perspective like me, by solving all potential issues I can foresee, I can maximize the compatibility of the program (including the lib it uses), not to break down on new servers time to time. That's what I did in the first two commits.

emersion (approved):
Thanks!

Issue #598, opened 2024-03-15
Support IMAP4 ID command (RFC 1971)


quzhi1:
Can you support IMAP ID command as specified in RFC 1971? https://www.rfc-editor.org/rfc/rfc2971.html

emersion:
Right… I've been holding off on that one because I think it's harmful: it enables clients and servers to implement implementation-specific workarounds instead of following the standard.

Can you explain your use-case?

quzhi1:
The use case is Yahoo. Yahoo advise users to use id command to identify themselves. Here is the quote:
> Client Indentification

> The Verizon Media IMAP server supports the IMAP ID command https://www.ietf.org/rfc/rfc2971.txt to gather statistics and facilitate troubleshooting. An IMAP client connecting to the Verizon Media servers should issue the ID command with the following attributes: NAME, VERSION, OS, OS-VERSION.

> NAME should be the partner name or the ID assigned during the approval process.
example: "name" "iPhone Mail"
"name" "com.android.email"
"name" "yahoo.com"

> VERSION is the version of that client. It's useful when watching for changes in behavior OS & OS-VERSION are additional info for determining client environment specific issues.

> A full ID command looks like this:
> C: a007 ID ("name" "<'client name'>" "version" "<'client version'>" "os" "<'client os'>" "os-version" "<'client os version'>")

https://developer.yahooinc.com/imap-smtp/

quzhi1:
It is also documented here: https://senders.yahooinc.com/developer/documentation/#imap-features

x0st:
@quzhi1 https://github.com/emersion/go-imap/pull/601

emersion:
Right, so it's a recommendation from Yahoo, but is there any real gain for clients which use the ID extension?

quzhi1:
> Right, so it's a recommendation from Yahoo, but is there any real gain for clients which use the ID extension?

@emersion Yes. I reached out to Yahoo, and they said if I set IMAP ID, they can detect the origin of the IMAP request, and can raise the rate limit. Also, if we do not pass the IMAP ID, they can potentially lock the Yahoo account if they see too many IMAP connections.

quzhi1:
This is the email Yahoo sent us. Basically they ask me to pass IMAP ID as part of their IMAP agreement:
[...]

PR #601, opened 2024-03-20
imapclient: support the id cmd

x0st:
I really need this in my project. Maybe you can consider merging this :)

### DIFF
--- a/id.go
+++ b/id.go
@@ -1,0 +1,5 @@
+package imap
+
+type IdData struct {
+	Data map[string]string
+}
--- a/imapclient/id.go
+++ b/imapclient/id.go
@@ -1,0 +1,100 @@
+package imapclient
+
+import (
+	"fmt"
+	"github.com/emersion/go-imap/v2"
+	"github.com/emersion/go-imap/v2/internal/imapwire"
+)
+
+func (c *Client) Id(keysAndValues ...string) *IdCommand {
+	if len(keysAndValues)%2 != 0 {
+		panic("imapclient: the length of keys and values is odd")
+	}
+
+	cmd := &IdCommand{}
+	enc := c.beginCommand("ID", cmd)
+
+	if len(keysAndValues) == 0 {
+		enc.SP().NIL()
+	} else {
+		enc.SP().Special('(')
+
+		for i, keyOrValue := range keysAndValues {
+			enc.Quoted(keyOrValue)
+
+			if i != len(keysAndValues)-1 {
+				enc.SP()
+			}
+		}
+
+		enc.Special(')')
+	}
+	enc.end()
+
+	return cmd
+}
+
+func (c *Client) handleId() error {
+	data, err := c.readId(c.dec)
+	if err != nil {
+		return fmt.Errorf("in id: %v", err)
+	}
+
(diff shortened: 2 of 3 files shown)


emersion:
Can you explain what your use-case is for this? Why is this a requirement for your project?

x0st:
I am copying the behaviour of an Android email app to make a bot. This is what the app does. So, I need to do this too

emersion:
Hm, sorry, not sure I'm following… Why is it important to behave exactly like the Android email app? Does the server reject your connection if you don't use the ID command? If so, which server is this?

x0st:
Unfortunately, I can't tell you the details. I am fine if you are against merging this PR.

emersion:
Until someone comes in with a solid use-case and explains how the ID command is actually useful for a reasonable scenario, I'd prefer to not add support for it.

quzhi1:
Can we merge this PR? 🙏  This can solve https://github.com/emersion/go-imap/issues/598

id.go
```diff
@@ -0,0 +1,5 @@
+package imap
+
+type IdData struct {
+	Data map[string]string
```
emersion, line 4: Can we use struct fields for the stuff defined in the RFC?

https://www.rfc-editor.org/rfc/rfc2971.html#section-3.3

imapclient/id.go
```diff
@@ -0,0 +1,100 @@
...
+	return cmd
+}
+
+func (c *Client) handleId() error {
+	data, err := c.readId(c.dec)
+	if err != nil {
+		return fmt.Errorf("in id: %v", err)
+	}
+
+	cmd := c.findPendingCmdFunc(func(cmd command) bool {
```
emersion, line 43: I think we can use `findPendingCmdByType` here?

imapclient/id.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func (c *Client) readId(dec *imapwire.Decoder) (*imap.IdData, error) {
+	var data = imap.IdData{Data: map[string]string{}}
+
+	if !dec.ExpectSP() {
+		return nil, dec.Err()
+	}
+
+	if !dec.ExpectNIL() {
```
emersion, line 67: `Expect*` will set the error field in the decoder if the server doesn't send NIL here. We shouldn't set the error field, since we accept either NIL or a list.

Would be easier to read with an early return:

```go
if dec.NIL() {
    return &data, nil
}
```

imapclient/id.go
```diff
@@ -0,0 +1,100 @@
...
+
+	return &data, nil
+}
+
+type IdCommand struct {
+	cmd
+	data imap.IdData
+}
+
+func (r *IdCommand) Data() imap.IdData {
```
emersion, line 98: Should probably be named `Wait()` for consistency with other commands, and return both an error and the data?

quzhi1:
@x0st Can you address the comment? 🙏

id.go
x0st: @emersion No

emersion: You're going to have to give me more details if you want me to understand your reply. In other words, please stick to constructive comments.

emersion:
Superseded by #621

Issue #602, opened 2024-03-22
Idle client v2 stops receiving new emails in some time.


serhiynovos:
I need a service to listening for incoming emails and send them to internal server. It's working but I noticed that in some time like 1 2 days it stops receiving a new emails. In console I don't see any errors. Is there are any way I can check idle client status and if it's failed I can restart client or even just fail client so container with it will be restarted automatically ?

```
package client

import (
	"bytes"
	"context"
	"fmt"
	"log"
	"sync"
	"text/template"

	"github.com/emersion/go-imap/v2"
	"github.com/emersion/go-imap/v2/imapclient"
	"github.com/mnako/letters"
	"mycom.imap.client/pkg/config"
	"mycom.imap.client/pkg/models"
	"mycom.imap.client/pkg/parsers"
	mycomserver "mycom.imap.client/pkg/mycom-server"
)

type IMAPClient struct {
	idleClient *imapclient.Client
	ctx        context.Context
	wg         *sync.WaitGroup
	idleCmd    *imapclient.IdleCommand
	config     config.Listener
}

func (client *IMAPClient) Start() {
	fmt.Printf("Start %s listener\n", client.config.Name)
	client.wg.Add(1)

	idleC, err := imapclient.DialTLS(client.config.Imap.Host, &imapclient.Options{
		UnilateralDataHandler: &imapclient.UnilateralDataHandler{
			Expunge: func(seqNum uint32) {
				log.Printf("message %v has been expunged", seqNum)
			},
			Mailbox: func(data *imapclient.UnilateralDataMailbox) {
				if data.NumMessages != nil {
					client.fetch(*data.NumMessages)
				}
			},
		},
	})

	if err != nil {
		log.Fatalf("failed to dial IMAP server: %v", err)
	}

	client.idleClient = idleC

	if err := idleC.Login(client.config.Imap.Username, client.config.Imap.Password).Wait(); err != nil {
		log.Fatalf("failed to login: %v", err)
	}

	if _, err := idleC.Select("INBOX", nil).Wait(); err != nil {
		log.Fatalf("failed to select INBOX: %v", err)
	}

	// Start idling
	idleCmd, err := idleC.Idle()
	if err != nil {
		log.Fatalf("IDLE command failed: %v", err)
	}
	client.idleCmd = idleCmd

	<-client.ctx.Done()
	client.Close()
}

func (client *IMAPClient) fetch(number uint32) {
	fmt.Println("Fetch", number)
[...]

PR #603, opened 2024-03-28
imapclient: Enhancement - Add custom interval to IDLE command

birudeghi:
Although RFC states an IDLE timeout of 29 minutes, many IMAP servers timeout much earlier than that. Other forums point to around 10-13 minutes as a 'sweet spot'.

This PR gives more flexibility for those who need to adjust the restart interval. This also resolves issue #602

### DIFF
--- a/imapclient/idle.go
+++ b/imapclient/idle.go
@@ -115,6 +115,9 @@ func (cmd *IdleCommand) runWithInterval(c *Client, child *idleCommand, restartIn
 			if cmd.err = child.Close(); cmd.err != nil {
 				return
 			}
+			if cmd.err = child.Wait(); cmd.err != nil {
+				return
+			}
 			if child, cmd.err = c.idle(); cmd.err != nil {
 				return
 			}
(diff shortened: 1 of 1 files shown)


emersion:
> many IMAP servers timeout much earlier than that

Out of curiosity, do you have any examples of this?

> Other forums point to around 10-13 minutes as a 'sweet spot'.

Any references?

---

Generally I'd prefer for `Idle` to Do The Right Thing™ by default. In other words, if in practice a restart interval of 10 minutes works better, I'm not opposed to lowering the current timeout.

If there really is a need to allow the restart interval to be customized, I'd prefer to break the API and add an `IdleOptions` struct, rather than introduce a completely new function (which BTW is a copy-paste of the other one).

birudeghi:
Lark IDLE timeout after every 10 minutes:

Screenshot 2024-04-10 at 20 17 34

Your change (https://github.com/emersion/go-imap/commit/ca0ddb75a697f13e66376b7f7be82cd13efdd507) does its job, albeit having to implement reconnect logic ourselves

PR #604, opened 2024-03-29
Add idle refresh fix

birudeghi:


emersion:
How does this relate to #603? It seems like these are exactly the same changes?

birudeghi:
Same changes, i'll close this one.

PR #620, opened 2024-06-05
imapclient: return greeting text on WaitGreeting

opsxolc:


opsxolc:
I've also added Templates mailbox attribute here, I didn't find any corresponding RFC, but it's commonly used anyways. If it's not ok to pull this changes together, I can move this part to a separate PR.

emersion:
Can you explain what the use-case is for getting the server greeting? Usually it's some arbitrary text that should be ignored, much like the text in the command replies.

opsxolc:
We use it to determine possible server soft such as axigen, zimbra, qqmail, courier, msexchange etc. And based on this information we may change the behavior of our service in some corner cases.

emersion:
Wouldn't the ID extension be a more appropriate way to do this?

emersion:
Closing because I don't think this feature is desirable.

PR #621, opened 2024-06-11
Support ID command

quzhi1:
Hi @emersion ,

I based my PR on https://github.com/emersion/go-imap/pull/601, and addressed all your comments in this one. Really appreciate your review 🙏 

I have one question though... The response of `c.Id(...).Wait()` is somehow always empty. I am not sure why the library can't read the response from server. Can you help me figure out why?

Great thanks!

### DIFF
--- a/id.go
+++ b/id.go
@@ -1,0 +1,15 @@
+package imap
+
+type IDData struct {
+	Name        string
+	Version     string
+	OS          string
+	OSVersion   string
+	Vendor      string
+	SupportURL  string
+	Address     string
+	Date        string
+	Command     string
+	Arguments   string
+	Environment string
+}
--- a/imapclient/id.go
+++ b/imapclient/id.go
@@ -1,0 +1,163 @@
+package imapclient
+
+import (
+	"fmt"
+
+	"github.com/emersion/go-imap/v2"
+	"github.com/emersion/go-imap/v2/internal/imapwire"
+)
+
+// ID sends an ID command.
+//
+// The ID command is introduced in RFC 2971. It requires support for the ID
+// extension.
+//
+// An example ID command:
+//
+//	ID ("name" "go-imap" "version" "1.0" "os" "Linux" "os-version" "7.9.4" "vendor" "Yahoo")
+func (c *Client) ID(idData *imap.IDData) *IDCommand {
+	cmd := &IDCommand{}
+	enc := c.beginCommand("ID", cmd)
+
+	if idData == nil {
+		enc.SP().NIL()
+		enc.end()
+		return cmd
+	}
+
+	enc.SP().Special('(')
+	isFirstKey := true
+	if idData.Name != "" {
+		addIDKeyValue(enc, &isFirstKey, "name", idData.Name)
+	}
+	if idData.Version != "" {
+		addIDKeyValue(enc, &isFirstKey, "version", idData.Version)
+	}
+	if idData.OS != "" {
+		addIDKeyValue(enc, &isFirstKey, "os", idData.OS)
+	}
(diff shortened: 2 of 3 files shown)


imapclient/id.go
```diff
@@ -0,0 +1,117 @@
...
+	return &data, nil
+}
+
+type IdCommand struct {
+	cmd
+	data imap.IdData
+}
+
+func (r *IdCommand) Wait() imap.IdData {
+	return r.data
```
emersion: Need to call `r.cmd.Wait()` here (and return the error if any)

imapclient/id.go
```diff
@@ -0,0 +1,117 @@
+package imapclient
+
+import (
+	"fmt"
+
+	"github.com/emersion/go-imap/v2"
+	"github.com/emersion/go-imap/v2/internal/imapwire"
+)
+
+func (c *Client) Id(keysAndValues ...string) *IdCommand {
```
emersion: Can we take `*IdData` as input here instead of the string list?

imapclient/id.go
```diff
@@ -0,0 +1,117 @@
...
+
+	currKey := ""
+	err := dec.ExpectList(func() error {
+		var keyOrValue string
+		if !dec.String(&keyOrValue) {
+			return fmt.Errorf("in id key-val list: %v", dec.Err())
+		}
+
+		if currKey == "" {
+			currKey = keyOrValue
```
emersion, line 112: Nit: an early return could be used here to de-indent the rest of the function.

```go
if currKey == "" {
    return nil
}

switch currKey {
…
}
```

imapclient/id.go
quzhi1: Ah that's why I can't get the response. This explains everything. Thank you so much!

quzhi1 (commented):
@emersion Thanks for reviewing my pull request. I addressed your comment. The PR is looking much cleaner now.

quzhi1:
@emersion Can you take a look again? 🙏 The PR is updated

id.go
```diff
@@ -0,0 +1,15 @@
+package imap
+
+type IdData struct {
```
emersion: Nit: can we use the proper upper-casing for things like "ID", "OS", "URL"?

See https://go.dev/wiki/CodeReviewComments#initialisms

(The "ID" thing applies to all symbols: `IDData`, `IDCommand`, `Client.ID`, and so on)

imapclient/id.go
```diff
@@ -0,0 +1,167 @@
...
+		enc.SP().NIL()
+		enc.end()
+		return cmd
+	}
+
+	enc.SP().Special('(')
+	isFirstKey := true
+	if idData.Name != "" {
+		addIdKeyValue(enc, isFirstKey, "name", idData.Name)
+		isFirstKey = false
```
emersion: Maybe the address of `isFirstKey` could be passed to `addIdKeyValue` to avoid repetition and always set the value to `true` at the end of the call. Alternatively, build a `[]string` of key-values and encode them in one loop.

emersion (commented):
Functionality looks good! Here are two final suggestions.

quzhi1:
@emersion Thanks for another round of review. The PR is updated. Can you take a look again?

emersion (approved):
Thank you!

quzhi1:
@emersion Can you merge it? I don't have permission to merge this pull request...

emersion:
Sorry, I wanted to cleanup the commit history and reword/reformat the docs a bit, but I got interrupted. Merging now!

quzhi1:
Thank you! This really means a lot! Really appreciate the help 👍 👍 👍

PR #627, opened 2024-07-16
Support complete section specification for BINARY, BINARY.PEEK, and BINARY.SIZE fetch data items

agcom:
Added the same fetch section specification fields in the `imap.FetchItemBodySection` to the `imap.FetchItemBinarySection` and `imap.FetchItemBinarySectionSize` structs (allowed by the RFC9051):

```go
type FetchItemBodySection struct {
	Peek bool

	Part []int

	Specifier       PartSpecifier
	HeaderFields    []string
	HeaderFieldsNot []string

	Partial *SectionPartial
}

type FetchItemBinarySection struct {
	Peek bool

	Part []int

	Specifier       PartSpecifier // NEW
	HeaderFields    []string      // NEW
	HeaderFieldsNot []string      // NEW

	Partial *SectionPartial
}

type FetchItemBinarySectionSize struct {
	Part []int

	Specifier       PartSpecifier // NEW
	HeaderFields    []string      // NEW
	HeaderFieldsNot []string      // NEW
}
```

Then, did the cascading changes required in `imapclient` and `imapserver` to read and write those new fields; reused the fetch section specification read/write logic of `imap.FetchItemBodySection`, by defining a private `fetchSectionSpec` proxy struct.

I tried to make no major/breaking API changes, and that's why it may seem like a mess.

### DIFF
--- a/fetch.go
+++ b/fetch.go
@@ -57,24 +57,37 @@ type SectionPartial struct {
 //
 //	imap.FetchItemBodySection{Specifier: imap.PartSpecifierHeader}
 type FetchItemBodySection struct {
+	Peek bool
+
+	Part []int
+
 	Specifier       PartSpecifier
-	Part            []int
 	HeaderFields    []string
 	HeaderFieldsNot []string
-	Partial         *SectionPartial
-	Peek            bool
+
+	Partial *SectionPartial
 }
 
 // FetchItemBinarySection is a FETCH BINARY[] data item.
 type FetchItemBinarySection struct {
-	Part    []int
+	Peek bool
+
+	Part []int
+
+	Specifier       PartSpecifier
+	HeaderFields    []string
+	HeaderFieldsNot []string
+
 	Partial *SectionPartial
-	Peek    bool
 }
 
 // FetchItemBinarySectionSize is a FETCH BINARY.SIZE[] data item.
 type FetchItemBinarySectionSize struct {
 	Part []int
+
+	Specifier       PartSpecifier
+	HeaderFields    []string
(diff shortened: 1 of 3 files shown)


emersion:
I don't believe RFC 9051 allows these. The ABNF reads:

```
section         = "[" [section-spec] "]"
section-binary  = "[" [section-part] "]"

section-spec    = section-msgtext / (section-part ["." section-text])
section-part    = nz-number *("." nz-number)
```

agcom:
> I don't believe RFC 9051 allows these. The ABNF reads:
> 
> ```
> section         = "[" [section-spec] "]"
> section-binary  = "[" [section-part] "]"
> 
> section-spec    = section-msgtext / (section-part ["." section-text])
> section-part    = nz-number *("." nz-number)
> ```

Ah, interesting. I didn't lookup the grammar, but just the explanation which doesn't seem to differentiate between `section` and `section-binary`.

> 6.4.5.1.  FETCH Section Specification
> 
>    Several FETCH data items reference "section" or "section-binary".
>    The section specification is a set of zero or more part specifiers
>    delimited by periods.  A part specifier is either a part number or
>    one of the following: HEADER, HEADER.FIELDS, HEADER.FIELDS.NOT, MIME,
>    and TEXT.  (Non-numeric part specifiers have to be the last specifier
>    in a section specification.)  An empty section specification refers
>    to the entire message, including the header.

Besides that, the IMAP server that I'm interacting with (probably Dovecot), does accept `BINARY[section]` and `BINARY.SIZE[section]`.

But I guess you're right; let's close this.

PR #629, opened 2024-07-23
Fixup fetch BINARY.SIZE decode bug

agcom:
Add expect the open square bracket of its section specification.


emersion (approved):
Good catch, thanks!

Issue #639, opened 2024-09-09
Getting `in response: imapwire: expected CRLF, got "]"` error when listing mailboxes


quzhi1:
Hi @emersion ,

I am having trouble listing folders for an IMAP account. This is the IMAP settings:
- Host: imap.startmail.com
- Port: 993

If I use
```golang
mailboxes, err := c.List("", "*", nil).Collect()
```
I get this error
```
in response: imapwire: expected CRLF, got "]"
```

If I use openssl to list mailboxes directly, this is the response:
```
a LIST "" "*"
* LIST (\HasNoChildren \UnMarked) "." "Downtown Nampa"
* LIST (\HasNoChildren \UnMarked) "." "Restoration Systems"
* LIST (\HasNoChildren \UnMarked) "." Steidl
* LIST (\HasNoChildren \UnMarked) "." Stability
* LIST (\HasNoChildren \UnMarked) "." SHE
* LIST (\HasNoChildren \UnMarked) "." "SDI UT Holdings"
* LIST (\HasNoChildren \UnMarked) "." Revoice
* LIST (\HasNoChildren \UnMarked) "." "Rawson Technologies"
* LIST (\HasNoChildren \UnMarked) "." "QuickBooks Payments"
* LIST (\HasNoChildren \UnMarked) "." "Quality First"
* LIST (\HasNoChildren \UnMarked) "." QFC
* LIST (\HasNoChildren \UnMarked) "." "Potential Clients"
* LIST (\HasChildren \UnMarked) "." "Old Clients"
* LIST (\HasNoChildren \UnMarked) "." "Old Clients.Bright Beacon"
* LIST (\HasNoChildren \UnMarked) "." "Old Clients.AtoZ Construction"
* LIST (\HasNoChildren \UnMarked) "." "Old Clients.Cabrillo"
* LIST (\HasNoChildren \UnMarked) "." "Old Clients.ProGear (K-Mac)"
* LIST (\HasNoChildren \UnMarked) "." "Old Clients.Affordable Mirror"
* LIST (\HasNoChildren \UnMarked) "." "Old Clients.Synthepor"
* LIST (\HasNoChildren \UnMarked) "." "Old Clients.Dallmann"
* LIST (\HasNoChildren \UnMarked) "." "Old Clients.True Lawn Care"
* LIST (\HasNoChildren \UnMarked) "." "Old Clients.The Well"
* LIST (\HasNoChildren \UnMarked) "." "Old Clients.West to East"
[...]

PR #640, opened 2024-09-10
Allow special characters in unquoted mailbox names

quzhi1:
Hi @emersion ,

Can I make this change to allow special character in unquoted mailbox names? IMAP providers such as StartMail does that.

I created an issue here: https://github.com/emersion/go-imap/issues/639. Then I found out may be we can do this change to solve the problem. Do you think my change is sufficient?

I used a startmail account to create a bunch of folders with special characters, and then verified my changes. It is working. Here is the related output while using `DebugWriter: os.Stdout` option:

```
T2 LIST "" "*"
* LIST (\HasNoChildren) "." Special#Hash
* LIST (\HasNoChildren) "." "Special%Percentage"
* LIST (\HasNoChildren) "." "(Parenthesis)"
* LIST (\HasNoChildren \Marked) "." [Special]
* LIST (\HasNoChildren \UnMarked \Sent) "." "Sent Messages"
* LIST (\HasNoChildren \UnMarked \Trash) "." "Deleted Messages"
* LIST (\HasNoChildren \UnMarked \Junk) "." Junk
* LIST (\HasNoChildren \UnMarked \Drafts) "." Drafts
* LIST (\HasNoChildren) "." INBOX
T2 OK List completed (0.001 + 0.000 secs).
```

Zhi Qu

### DIFF
--- a/internal/imapwire/decoder.go
+++ b/internal/imapwire/decoder.go
@@ -399,8 +399,9 @@ func (dec *Decoder) ExpectAString(ptr *string) bool {
 	if dec.Literal(ptr) {
 		return true
 	}
-	// TODO: accept unquoted resp-specials
-	return dec.ExpectAtom(ptr)
+	// We cannot do dec.Atom(ptr) here because sometimes mailbox names are unquoted,
+	// and they can contain special characters like `]`.
+	return dec.ExpectText(ptr)
 }
 
 func (dec *Decoder) String(ptr *string) bool {
--- a/internal/imapwire/decoder.go
+++ b/internal/imapwire/decoder.go
@@ -27,9 +27,9 @@ func IsAtomChar(ch byte) bool {
 }
 
 // Is non-empty char
-func IsNonEmptyChar(ch byte) bool {
+func IsNonEmptyOrParenthesisChar(ch byte) bool {
 	switch ch {
-	case ' ', '\r', '\n':
+	case ' ', '\r', '\n', '(', ')':
 		return false
 	default:
 		return !unicode.IsControl(rune(ch))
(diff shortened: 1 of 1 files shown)


quzhi1:
@emersion Can you take a look at this proposed change? I am not sure what I did is correct 😊

emersion:
Thanks for the PR! Could we maybe introduce a new `isAStringChar` helper function, which would work like `IsAtomChar` but follow the `ASTRING-CHAR` rules? It could return `IsAtomChar(ch) || ch == ']'`:

```abnf
ASTRING-CHAR   = ATOM-CHAR / resp-specials
resp-specials   = "]"
```

Then we can use

```go
return dec.Expect(dec.Func(ptr, isAStringChar), "ASTRING-CHAR")
```

at the end of `ExpectAString`.

Does this make sense?

After reading the diff, it seems like your changes are functionally equivalent, but I'd prefer to structure the code the way indicated above to follow more closely the RFC (in terms of names and logic) and have less unused exported functions.

quzhi1:
@emersion Thanks for the review!
I agree we should introduce `isAStringChar` helper function.
I took a look at the definition of `ASTRING-CHAR` in RFC 9051, I found the `%` and `*` characters also needs special casing. I updated this PR. Is the change looking good now?

emersion:
> I found the % and * characters also needs special casing

That sounds weird to me... My reading of the RFC is that they shouldn't need special casing.

quzhi1:
@emersion You are right... I misread the `resp-specials` definition. It is only `]`. `*` and `%` are not included. I just updated my PR. Can you take a look again?

emersion (approved):
LGTM, thank you!

PR #648, opened 2024-11-10
Allow returning SORT capability

tgulacsi:
Please let SORT through.


emersion:
This is not enough for SORT support. The server side is unimplemented right now.

tgulacsi:
I've found it right after the first try (Unknown command).
Sorry for the noise.

PR #652, opened 2024-11-20
Allow continuous SP between CAPABILITY items

qjebbs:
Before logged in, imap-n.global-mail.cn sends:

    * CAPABILITY IMAP4<...>LITERAL+<SP><SP>STARTTLS<...>SPECIAL-USE

PR #653, opened 2024-11-22
Allow continuous SP for all commands

qjebbs:
Before logged in, imap-n.global-mail.cn sends:

    * CAPABILITY IMAP4<...>LITERAL+<SP><SP>STARTTLS<...>SPECIAL-USE


qjebbs:
This is another solution for the same problem as #652, by not adding `for dec.SP() {}` everywhere, solving all these kind of problems.

emersion:
I prefer to not add broad workarounds like these, so that we stay as close to the RFC as possible, and only relax our parsing when absolutely necessary.

PR #654, opened 2024-11-22
Make commands serial to comply RFC 3501 section-5.5

### LINKED ISSUE #651: CAPABILITY Interferes with LOGIN Command
Backgroud:

1. When the imapclient starts, it sends a CAPABILITY command to the server if the server greets it without CAPABILITY. The command is executed in another goroutine.
2. Some server (`imaphz.qiye.163.com` here) seems to ignore new commands when the current command is still being processed.

UPDATE:

`imaphz.qiye.163.com` is not the one to blame, the [RFC 3501](https://datatracker.ietf.org/doc/html/rfc3501#section-5.5) specification says:

```
Therefore, 
if the client sends any command other than FETCH, STORE, or SEARCH, it
MUST wait for the completion result response before sending a command
with message sequence numbers.
```

Consider the following code:

```go
// conn is a net.Conn to imaphz.qiye.163.com
c = imapclient.New(conn, &imapclient.Options{
    DebugWriter: os.Stdout,
})
c.WaitGreeting()
err := c.Login(userName, password).Wait()
if  err != nil {
	return nil, fmt.Errorf("login %q: %w", userName, err)
}
```

The `LOGIN` and  background spawned `CAPABILITY` are executed in different goroutines. The order of the commands is not guaranteed. So I see a random failure when initializing the imapclient.

1. If the `CAPABILITY` command is executed first, and the `LOGIN` command is executed before the `CAPABILITY` command finishes, the server seems to ignore the `LOGIN` command. The command receives nothing and reports an unexpected EOF error after 60 seconds.

    ```
    * OK IMAP4 ready
    T1 CAPABILITY
    T2 LOGIN "xxx@xxx.com" "xxxx"
    * CAPABILITY IMAP4rev1 XAPPLEPUSHSERVICE XLIST SPECIAL-USE AUTH=PLAIN ID
    T1 OK completed
    --- FAIL: TestIMAPLogin (60.13s)
        imap_test.go:57: login "xxx@xxx.com": unexpected EOF
    FAIL
    ```

1. If the `LOGIN` command happens to be executed first, the server will respond. (`CAPABILITY` command is ignored until timeout. Though the client initialized with success)

    ```
    * OK IMAP4 ready
    T1 LOGIN "xxx@xxx.com" "xxxx"
    T2 CAPABILITY
    T1 NO ERR.LOGIN.PASSERR
    T3 LOGOUT
    --- FAIL: TestIMAPLogin (0.20s)
        imap_test.go:57: login "xxx@xxx.com": imap: NO ERR.LOGIN.PASSERR
    ```

~~Though It's the server's fault, I can't change the server's behavior. Is there a way to fix this issue? Maybe to lock up imapclient before a background command finishes (not accept new ones)?~~

## Workaround

UPDATE: In theory, there is no workaround to this problem

qjebbs:
Fixes #651

### DIFF
--- a/imapclient/client.go
+++ b/imapclient/client.go
@@ -21,6 +21,7 @@ package imapclient
 
 import (
 	"bufio"
+	"context"
 	"crypto/tls"
 	"errors"
 	"fmt"
@@ -35,6 +36,7 @@ import (
 	"github.com/emersion/go-imap/v2"
 	"github.com/emersion/go-imap/v2/internal"
 	"github.com/emersion/go-imap/v2/internal/imapwire"
+	"github.com/emersion/go-imap/v2/internal/serial"
 )
 
 const (
@@ -76,6 +78,10 @@ type Options struct {
 	UnilateralDataHandler *UnilateralDataHandler
 	// Decoder for RFC 2047 words.
 	WordDecoder *mime.WordDecoder
+	// CommandsQueueSize is the size of the queue for commands that are waiting
+	// to be sent in order to the server. The order of commands exceeding this
+	// size is not guaranteed. The default size is 16.
+	CommandsQueueSize int
 }
 
 func (options *Options) wrapReadWriter(rw io.ReadWriter) io.ReadWriter {
@@ -145,6 +151,8 @@ type Client struct {
 	decCh  chan struct{}
 	decErr error
 
+	serial *serial.Control
+
 	mutex        sync.Mutex
 	state        imap.ConnState
 	caps         imap.CapSet
@@ -166,6 +174,9 @@ func New(conn net.Conn, options *Options) *Client {
 	if options == nil {
 		options = &Options{}
 	}
+	if options.CommandsQueueSize <= 0 {
+		options.CommandsQueueSize = 16
+	}
 
 	rw := options.wrapReadWriter(conn)
 	br := bufio.NewReader(rw)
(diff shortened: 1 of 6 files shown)


imapclient/client.go
```diff
@@ -166,6 +174,9 @@ func New(conn net.Conn, options *Options) *Client {
 	if options == nil {
 		options = &Options{}
 	}
+	if options.CommandsQueueSize <= 0 {
```
qjebbs, line 177: Remove this if this is acceptable: by default, commands order is not guaranteed.

imapclient/client.go
```diff
@@ -166,6 +174,9 @@ func New(conn net.Conn, options *Options) *Client {
...
+		options.CommandsQueueSize = 16
+	}
```
qjebbs, line 179: Remove this if this is acceptable: by default, commands order is not guaranteed.

Benefit: reduce 1 goroutine for each client.

imapclient/client.go
```diff
@@ -1190,6 +1201,38 @@ func (cmd *commandBase) wait() error {
...
+func (cmd *commandBase) preStart(serial *serial.Control, name string) {
+	if !isSerial(name) {
+		return
+	}
+	_, cancel := serial.New(context.Background())
+	cmd.cancel = cancel
+	// If the *Client has been closed, commands in the queue will still be executed
+	// but report 'use of closed network connection' error.
+
+	// TODO: cancel commands in queue if *Client (also the *serial.Control) is closed.
```
qjebbs, line 1213: It requires a lot of change.

PR #655, opened 2024-12-04
fetch mails with time 

brigisroy:
Updated the existing functionality fetch only with date to fetch with time


emersion:
[RFC 9051 section 6.4.4](https://www.rfc-editor.org/rfc/rfc9051.html#section-6.4.4) says the time and timezone are disregarded. The grammar has:

```abnf
date            = date-text / DQUOTE date-text DQUOTE
date-text       = date-day "-" date-month "-" date-year
```

So, I don't think this is correct.

PR #656, opened 2024-12-07
Support IMAP server response token case-insensitively

quzhi1:
Hello @emersion ,

I saw some IMAP server are returning tags in different cases. For example:
```
a select INBOX
* 2 EXISTS
* 2 Recent
* FLAGS (\Deleted \Seen \Flagged \Answered $Forwarded)
* OK [UIDVALIDITY 1690369825] UIDs valid
* OK [PERMANENTFLAGS (\Deleted \Seen \Flagged \Answered $Forwarded)] Permanent flags
a OK [READ-WRITE] SELECT completed
```

Note the `* 2 Recent` line. The `Recent` keyword is not all upper-cased.

In RFC-9051 IMAP protocol, it seems like client should parse server tags without case-sensitivity.

> Unless otherwise noted, all alphabetic characters are case insensitive. The use of uppercase or lowercase characters to define token strings is for editorial clarity only. Implementations MUST accept these strings in a case-insensitive fashion.

https://www.rfc-editor.org/rfc/rfc9051.html#section-9

This PR will start accepting such tags. Can you take a look?


emersion:
Sounds like a good idea to me, but can we move the capitalization to the caller (inside `readResponse()`), so that `typ` is always compared case-insensitively?

quzhi1:
@emersion Thanks for the reply! I tried to move it up, but `typ` variable changes its value in this line: https://github.com/emersion/go-imap/blob/v2/imapclient/client.go#L794. So even if we do `strings.ToUpper(typ)` in the caller, it will still be overwritten.
I updated my pull request to convert `typ` variable in both places. Can you take a look again?

imapclient/client.go
```diff
@@ -796,7 +800,8 @@ func (c *Client) readResponseData(typ string) error {
 		}
 	}
 
-	switch typ {
+	// All response type are case insensitive
+	switch strings.ToUpper(typ) {
```
emersion, line 804: Since `typ` is now upper-cased in the caller, I don't think we still need that one?

emersion: Oh, just read your comment. Makes sense!

emersion (approved):
Thanks!

quzhi1:
@emersion I don't have permission to merge this PR. Can you merge it? 🙇

PR #673, opened 2025-03-03
imapclient: use dialer in DialInsecure

snekROmonoro:
The default `net` package dialer doesn't have timeout. The dialer defined and used in client connection does.


emersion (approved):
Thanks!

PR #674, opened 2025-03-03
imapserver: Support RECENT for IMAP4rev1 servers

foxcpp:
`\Recent` flag is used by several IMAP4rev1 clients including Thunderbird and aerc.

1. STATUS data item: RECENT.
2. SELECT untagged response: RECENT.
3. Removed hacks that add `RECENTS 0` responses.
4. Conn.IsEnabled(Cap) method

I added Conn.IsEnabled method for checking whether a certain extension is enabled. This can be used by backend implementations to disable/enable (potentially expensive) RECENT handling if IMAP4rev2 is in use.

### DIFF
--- a/imapserver/conn.go
+++ b/imapserver/conn.go
@@ -85,6 +85,13 @@ func (c *Conn) Bye(text string) error {
 	return closeErr
 }
 
+func (c *Conn) EnabledCaps() imap.CapSet {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	return c.enabled.Copy()
+}
+
 func (c *Conn) serve() {
 	defer func() {
 		if v := recover(); v != nil {
@@ -583,6 +590,14 @@ func (w *UpdateWriter) WriteExpunge(seqNum uint32) error {
 // WriteNumMessages writes an EXISTS response.
 func (w *UpdateWriter) WriteNumMessages(n uint32) error {
 	return w.conn.writeExists(n)
+}
+
+// WriteNumRecent writes an RECENT response (not used in IMAP4rev2, will be ignored).
+func (w *UpdateWriter) WriteNumRecent(n uint32) error {
+	if w.conn.enabled.Has(imap.CapIMAP4rev2) || !w.conn.server.options.caps().Has(imap.CapIMAP4rev1) {
+		return nil
+	}
+	return w.conn.writeObsoleteRecent(n)
 }
 
 // WriteMailboxFlags writes a FLAGS response.
--- a/imapserver/status.go
+++ b/imapserver/status.go
@@ -14,13 +14,10 @@ func (c *Conn) handleStatus(dec *imapwire.Decoder) error {
 	}
 
 	var options imap.StatusOptions
-	recent := false
 	err := dec.ExpectList(func() error {
-		isRecent, err := readStatusItem(dec, &options)
+		err := readStatusItem(dec, &options)
 		if err != nil {
 			return err
-		} else if isRecent {
-			recent = true
 		}
 		return nil
 	})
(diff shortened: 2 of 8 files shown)


imapserver/status.go
```diff
@@ -110,12 +107,12 @@ func readStatusItem(dec *imapwire.Decoder, options *imap.StatusOptions) (isRecen
 	case "DELETED-STORAGE":
 		options.DeletedStorage = true
 	case "RECENT":
-		isRecent = true
+		options.NumRecent = true
```
emersion, line 117: Can we return a failure if the server hasn't advertised IMAP4rev1? (ie, it's a pure IMAP4rev2 server)

select.go
```diff
@@ -16,6 +16,8 @@ type SelectData struct {
 	PermanentFlags []Flag
 	// Number of messages in this mailbox (aka. "EXISTS")
 	NumMessages uint32
+	// Number of recent messages in this mailbox ("RECENT") (IMAP4rev1 only).
```
emersion: Could we mention "Obsolete, IMAP4rev1 only" to make it clear that it's not great to use it even on IMAP4rev1?

status.go
```diff
@@ -3,6 +3,7 @@ package imap
 // StatusOptions contains options for the STATUS command.
 type StatusOptions struct {
 	NumMessages bool
+	NumRecent   bool // IMAP4rev1 only
```
emersion: Ditto

imapserver/conn.go
```diff
@@ -85,6 +85,13 @@ func (c *Conn) Bye(text string) error {
 	return closeErr
 }
 
+func (c *Conn) IsEnabled(cap imap.Cap) bool {
```
emersion: What is the purpose of exporting this method?

(If it's useful, maybe we should export an `EnabledCaps` method instead which would cover more use-cases… But need to be careful around locks.)

imapserver/conn.go
```diff
@@ -585,6 +592,14 @@ func (w *UpdateWriter) WriteNumMessages(n uint32) error {
 	return w.conn.writeExists(n)
 }
 
+// WriteNumRecent writes an RECENT response (not used in IMAP4rev2, will be ignored).
+func (w *UpdateWriter) WriteNumRecent(n uint32) error {
+	if w.conn.IsEnabled(imap.CapIMAP4rev2) {
```
emersion: Probably should be `|| !w.conn.server.options.caps.Has(imap.CapIMAP4rev1)`: if the server is IMAP4rev2-only, we shouldn't send RECENT even if the client hasn't explicitly enabled IMAP4rev2.

emersion (commented):
This is perhaps a bit confusing for the client-side: the `SelectData`, `StatusOptions` and `StatusData` fields are ignored there. Maybe we should at least document that these are server-only, and/or error out when used on the client side.

imapserver/conn.go
foxcpp: >This can be used by backend implementations to disable/enable (potentially expensive) RECENT handling if IMAP4rev2 is in use.

foxcpp:
>Maybe we should at least document that these are server-only, and/or error out when used on the client side.

Would it be better if I added support on the client-side as well?

emersion:
I'd rather not add client side support for this obsolete feature.

imapserver/conn.go
emersion: I'd prefer to either leave it for later, or return the whole cap set.

foxcpp:
Fair enough

imapserver/status.go
foxcpp: Done (handleStatus).

select.go
foxcpp: Done.

status.go
foxcpp: Done.

imapserver/conn.go
foxcpp: I changed it to EnabledCaps method that creates a copy of enabled set and returns it.

imapserver/conn.go
foxcpp: Done.

imapserver/conn.go
emersion: FWIW, the client side uses a copy-on-write approach (we can add this optimization later).

emersion (approved):
Thanks!

imapserver/conn.go
foxcpp: As an server optimization, it does not really seem to be necessary to call EnabledCaps more than once after first SELECT as clients are not allowed to issue ENABLE after it:
> Clients MUST NOT issue ENABLE once they SELECT/EXAMINE a mailbox
(RFC 5161)

Issue #574, opened 2023-12-19
Implement limits for parser


emersion:
- Recursion limits for BODYSTRUCTURE and SEARCH
- Size limit on command string size (something greater than 8K)

emersion:
https://github.com/emersion/go-imap/commit/f7899667eda72592d32fd54d851ffe6c4c1e9684 implements a recursion limit for lists in general.

PR #677, opened 2025-03-17
Implement ID properly

vzeroupper:
This change supports ID command in the way it's described in RFC. I've read the discussion in #598, #601 and #621.

Our case is we have plenty of legacy and we use IMAP as a message fetching/storing protocol between some backend services and have a requirement to pass IPs between those. We used to use v1 and now we would like to stay up-to-date with v2.

Google "strongly recommends" to use ID with fields that are currently impossible to send and receive, namely, `contact` and `remote-host`: https://developers.google.com/gmail/imap/imap-extensions#checking_for_the_presence_of_extensions

I understand that supporting someone's legacy or gmail's wishes is hardly a reason to include a feature. And while I'm unfamiliar with the problems ID can cause, I'd argue that if ID is implemented, it should be done according to RFC. Current implementation has the following mismatches:
1. Custom fields are not supported and the set of fields supported is listed in RFC as an example
2. An error on `NIL` values in response
3. No field case insensitivity support in response
4. No field/value pair count and literal length limit checks
5. No `ID NIL` request support and no distinction between `ID NIL` and `ID ()` in response

Please consider merging this.


emersion:
I've already said no to proprietary extensions in the past, and this isn't something that I want to change at this time.

> Custom fields are not supported

Correct, by design.

>  the set of fields supported is listed in RFC as an example

That's not my understanding. The RFC _defines_ some standard fields in section 3.3.

> An error on NIL values in response

Sounds like a bug indeed.

> No field case insensitivity support in response

Also sounds like a bug.

> No field/value pair count and literal length limit checks

This should be implemented at the parser level: https://github.com/emersion/go-imap/issues/574

> No ID NIL request support and no distinction between ID NIL and ID () in response

What is the difference between `NIL` and `()`?

vzeroupper:
> What is the difference between `NIL` and `()`?

I don't know because very little semantics is defined in the RFC. The syntax allows it and it's easy to do, so why force it upon users?

> This should be implemented at the parser level: #574

I don't think specific command limits need to be there, but I'm not very familiar with the decoder, maybe I'm missing something.

> That's not my understanding. The RFC defines some standard fields in section 3.3.

It doesn't matter how you see the fields defined, my point is the RFC says they're "available", not specified what it means, and doesn't go against any non-availables:
- 3.1 says "Fields are permitted to be any IMAP4 string, and values are permitted
   to be any IMAP4 string or NIL."
- 3.3 says "Any string may be sent as a field, but the following are defined to
   describe certain values that might be sent.  Implementations are free
   to send none, any, or all of these."
-

vzeroupper:
> What is the difference between `NIL` and `()`?

I don't know because very little semantics is defined in the RFC. The syntax allows it and it's easy to do, so why force it upon users?

> This should be implemented at the parser level: #574

I don't think specific command limits need to be there, but I'm not very familiar with the decoder, maybe I'm missing something.

> That's not my understanding. The RFC defines some standard fields in section 3.3.

It doesn't matter how you see the fields defined, my point is the RFC says they're "available", not specifying what it means, and doesn't go against any non-availables:
- 3.1 says "Fields are permitted to be any IMAP4 string"
- 3.3 says "Any string may be sent as a field, but the following are defined to
   describe certain values that might be sent. Implementations are free
   to send none, any, or all of these."
- 3.3 defines field length and field-value count limits

This shows a clear intent for implementations to go for any fields they want, and they do.

emersion:
> I don't know because very little semantics is defined in the RFC. The syntax allows it and it's easy to do, so why force it upon users?

If there's no practical difference, why add more complexity to the API?

> I don't think specific command limits need to be there, but I'm not very familiar with the decoder, maybe I'm missing something.

Ah, you meant the RFC limits. I'd rather not add these on the client side.

> my point is the RFC says they're "available", not specifying what it means, and doesn't go against any non-availables

Sure, but that's effectively proprietary APIs.

I did not intend to merge support for ID in the first place, and accepted it in part because vendor-specific fields were kept out of scope.

emersion:
Closing per the above discussion.

vzeroupper:
> If there's no practical difference, why add more complexity to the API?

It's how much of the RFC you're willing to give up in order to make the API simpler.

One explanation would be in 3.1 `The client may also send NIL instead of the list, indicating that it wants to send no information, but would still accept a server response.` I personally don't know in what use case it can be important to distinguish NIL from () and NIL from "".

> Ah, you meant the RFC limits. I'd rather not add these on the client side.

The deciding factor to whether add them or not was to have a clear error as servers may not respond with a well phrased message in case client sends something over the limits.

> Sure, but that's effectively proprietary APIs.

There's text describing how the command should and must be used in sections 3 and 7. Making a proprietary APIs out of IMAP with ID would be a violation of the RFC. But of course are no methods to enforce compliance.

> I did not intend to merge support for ID in the first place, and accepted it in part because vendor-specific fields were kept out of scope.

IMAP providers implement their features based on RFCs, and might not constrain themselves to the set of fields given there. It can't be assumed that they're necessarily misusing the command. If they (likely) aren't, they're as legitimate as those that do constrain themselves. And Yahoo doesn't even require ID to be sent to them, so why have the command at all if if you see it as such a slippery slope?

Issue #533, opened 2023-07-05
Extension support status


emersion:
See:

- Capabilities: https://www.iana.org/assignments/imap-capabilities/imap-capabilities.xhtml#imap-capabilities-1
- Active drafts: https://datatracker.ietf.org/doc/search?name=imap&sort=&activedrafts=on&by=group&group=
- https://gist.github.com/nevans/8ef449da0786f9d1cc7c8324a288dd9b

| Extension             |  RFC   | Client | Server |
|-----------------------|--------|--------|--------|
| IMAP4rev1             | [3501] | ✓      | ✓      |
| IMAP4rev2             | [9051] | ✓      | ✓      |
| NAMESPACE †           | [2342] | ✓      | ✓      |
| UNSELECT †            | [3691] | ✓      | ✓      |
| UIDPLUS †             | [4315] | ✓      | ✓      |
| ESEARCH †             | [4731] | ✓      | ✓      |
| SEARCHRES †           | [5182] | ✓      | ✓      |
| ENABLE †              | [5161] | ✓      | ✓      |
| IDLE †                | [2177] | ✓      | ✓      |
| SASL-IR †             | [4959] | ✓      | ✓      |
| LIST-EXTENDED †       | [5258] | ✓      | ✓      |
| LIST-STATUS †         | [5819] | ✓      | ✓      |
| MOVE †                | [6851] | ✓      | ✓      |
| LITERAL- †            | [7888] | ✓      | ✓      |
| STATUS=SIZE †         | [8438] | ✓      | ✓      |
| ACL                   | [4314] | ✗      | ✗      |
| APPENDLIMIT           | [7889] | ✓      | ✓      |
| BINARY                | [3516] | ✓      | ✓      |
| CATENATE              | [4469] | ✗      | ✗      |
| CHILDREN              | [3348] | ✓      | ✓      |
| CONDSTORE             | [7162] | ✓      | ✗      |
| CONVERT               | [5259] | ✗      | ✗      |
| CREATE-SPECIAL-USE    | [6154] | ✗      | ✗      |
| ESORT                 | [5267] | ✗      | ✗      |
| FILTERS               | [5466] | ✗      | ✗      |
| ID                    | [2971] | ✓      | ✗      |
| LANGUAGE              | [5255] | ✗      | ✗      |
| LIST-MYRIGHTS         | [8440] | ✗      | ✗      |
| LITERAL+              | [7888] | ✓      | ✗      |
[...]

PR #683, opened 2025-04-09
fixes for fetches 

dejanstrbac:
- wherever numSet is passed in session methods, also numKind is needed so the call can be interpreted right
- in fetch, we need WriteModSeq if options.ModSeq is set
- empty lists were wrongly converted to () instead of NIL atom, e.g. resulting in `("image" "png" () "<logo>" NIL "BASE64" 3608 NIL ("inline" ()) NIL NIL)` instead of `("image" "png" NIL "<logo>" NIL "base64" 3608 NIL ("inline" NIL) NIL NIL)`

### DIFF
--- a/imapserver/copy.go
+++ b/imapserver/copy.go
@@ -13,7 +13,7 @@ func (c *Conn) handleCopy(tag string, dec *imapwire.Decoder, numKind NumKind) er
 	if err := c.checkState(imap.ConnStateSelected); err != nil {
 		return err
 	}
-	data, err := c.session.Copy(numSet, dest)
+	data, err := c.session.Copy(numSet, numKind, dest)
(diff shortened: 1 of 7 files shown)


emersion:
> wherever numSet is passed in session methods, also numKind is needed so the call can be interpreted right

Not sure I understand: it should be possible to type-switch on a `NumSet` to figure out which kind it is.

> in fetch, we need WriteModSeq if options.ModSeq is set

CONDSTORE is not yet supported on the server side (https://github.com/emersion/go-imap/issues/533). I think server support requires more than just `WriteModSeq`?

> empty lists were wrongly converted to () instead of NIL atom

Good catch! The ABNF doesn't allow empty lists.

PR #684, opened 2025-04-14
Correct missing NIL atom instead of ()

dejanstrbac:
empty lists were wrongly converted to () instead of NIL atom, e.g. resulting in `("image" "png" () "<logo>" NIL "BASE64" 3608 NIL ("inline" ()) NIL NIL)` instead of` ("image" "png" NIL "<logo>" NIL "base64" 3608 NIL ("inline" NIL) NIL NIL)`


emersion (approved):
Thanks!

PR #685, opened 2025-05-08
Use correct formatting to wrap errors

Fizzadar:
Fixes error unwrapping, currently the nested errors are lost.

### DIFF
--- a/imapclient/acl.go
+++ b/imapclient/acl.go
@@ -64,7 +64,7 @@ func (cmd *GetACLCommand) Wait() (*GetACLData, error) {
 func (c *Client) handleMyRights() error {
 	data, err := readMyRights(c.dec)
 	if err != nil {
-		return fmt.Errorf("in myrights-response: %v", err)
+		return fmt.Errorf("in myrights-response: %w", err)
 	}
 	if cmd := findPendingCmdByType[*MyRightsCommand](c); cmd != nil {
 		cmd.data = *data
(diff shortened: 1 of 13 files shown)


emersion:
`%v` is used on purpose in most places. When `%w` is used, the wrapped errors essentially become part of the API. That is, changing an internal implementation may change wrapped errors and break users. As a result, I prefer to only expose underlying errors when there is a good reason to do so.

Fizzadar:
Interesting, I had not thought of error wrapping that way before. I see how the wrapped errors become part of the API, which could change depending on Go version/etc. But I'm not entirely sure how this has any user facing impact? Surly it would only matter if users were explicitly checking error types (as I am), in which case they are already aware of how error wrapping works..

In my specific case the aim is to identify which errors are network errors. Without proper wrapping this is impossible (short of string matching). Roughly I have:

```go
var neterr net.Error
if errors.As(err, &neterr) {
    isNetwork = true
}
...
```

No worries if you'd rather stick to the non-wrapped errors, I'll just merge into my fork and use that. Intrigued to know if other Go projects do/don't wrap errors for this same reason 🤔

emersion:
This topic is discussed here: https://go.dev/blog/go1.13-errors#whether-to-wrap

I'm not against wrapping when it makes sense, I'm just worried about wrapping by default without a second thought.

Fizzadar:
> This topic is discussed here: https://go.dev/blog/go1.13-errors#whether-to-wrap

Ah thank you for linking that, very helpful.

> I'm not against wrapping when it makes sense, I'm just worried about wrapping by default without a second thought.

Fair enough. For my "catch all network errors" use case that basically requires wrapping everything since the network sits at one end and the caller at the other, any errors along the call stack would need to be wrapped. I'll close the PR for now, I would be very interested to hear in alternatives to my wrap everything approach, I have yet to find one.

emersion:
Could you describe why you need to catch all network errors?

Fizzadar:
> Could you describe why you need to catch all network errors?

I'm working on a desktop email client and need to display errors in the UI. We'd like to show network errors separately such that they can be ignored (and non-network errors are shown as a "something broke bad restart the app" kind of thing).

PR #687, opened 2025-05-19
imapserver: add CONDSTORE support

### LINKED ISSUE #82: Support HIGHESTMODSEQ
Often, multiple IMAP (RFC 3501) clients need to coordinate changes to
   a common IMAP mailbox.  Examples include different clients working on
   behalf of the same user, and multiple users accessing shared
   mailboxes.  These clients need a mechanism to synchronize state
   changes for messages within the mailbox.  They must be able to
   guarantee that only one client can change message state (e.g.,
   message flags) at any time.  An example of such an application is use
   of an IMAP mailbox as a message queue with multiple dequeueing
   clients.

These days many clients work with the same mail server, your desktop, laptop, mobile devices are all retrieving mails and updating them. So, it would be safe to support this IMAP extension which can be used to avoid overwriting some other client's changes.

https://tools.ietf.org/html/rfc4551

dejanstrbac:
closes #82


imapserver/select.go
```diff
@@ -64,6 +97,12 @@ func (c *Conn) handleSelect(tag string, dec *imapwire.Decoder, readOnly bool) er
 		}
 	}
 
+	if data.HighestModSeq > 0 && (options.CondStore || c.enabled.Has(imap.CapCondStore)) {
```
emersion, line 100: The RFC says:

> This document adds two new response codes: HIGHESTMODSEQ and
   NOMODSEQ.  One of these two response codes MUST be returned in an OK
   untagged response for any successful SELECT/EXAMINE command issued
   after a CONDSTORE enabling command.
   
If we don't have a HIGHESTMODSEQ, we should send NOMODSEQ.

emersion, line 100: `options.CondStore` indicates whether the server must include MODSEQ in all FETCH responses: https://datatracker.ietf.org/doc/html/rfc7162#section-3.1.8

It's unrelated to the HIGHESTMODSEQ returned in the SELECT command. In other words, we should always send HIGHESTMODSEQ, even if `options.CondStore` isn't enabled.

imapserver/select.go
```diff
@@ -9,7 +9,41 @@ import (
...
+				if !dec.SP() {
+					break
+				}
+			}
+
+			if !dec.ExpectSpecial(')') {
+				return dec.Err()
+			}
+		} else {
+			return dec.Err()
```
emersion, line 42: At this point, `(` did not follow the SP, so this is invalid syntax. However, we didn't use the `Expect` family of functions, so `Err` will be nil. We should probably replace `dec.Special('(')` with:

```go
if !dec.ExpectSpecial('(') {
	return dec.Err()
}
```

However, it would be easier to use `ExpectList`.

imapserver/store.go
```diff
@@ -13,7 +14,27 @@ func (c *Conn) handleStore(dec *imapwire.Decoder, numKind NumKind) error {
...
+
+	options := imap.StoreOptions{}
+	if dec.Special('(') {
+		var param string
+		if !dec.ExpectAtom(&param) {
+			return dec.Err()
+		}
+
+		if strings.ToUpper(param) == "UNCHANGEDSINCE" {
+			if !dec.ExpectSP() || !dec.ExpectModSeq(&options.UnchangedSince) || !dec.ExpectSpecial(')') || !dec.ExpectSP() {
```
emersion, line 29: Nit: could we move the last part of the condition, `if !dec.ExpectSpecial(')') || !dec.ExpectSP() {`, below outside of the `if strings.ToUpper(param) == "UNCHANGEDSINCE" {`? It would apply to other store modifiers as well.

imapclient/client_test.go
```diff
@@ -102,6 +102,7 @@ func newMemClientServerPair(t *testing.T) (net.Conn, io.Closer) {
 		Caps: imap.CapSet{
 			imap.CapIMAP4rev1: {},
 			imap.CapIMAP4rev2: {},
+			imap.CapCondStore: {}, // Add CONDSTORE capability
```
emersion, line 105: Nit: can this comment be dropped? I don't think it adds any useful info which is not already obvious with the code.

imapserver/capability.go
```diff
@@ -75,6 +75,7 @@ func (c *Conn) availableCaps() []imap.Cap {
 				imap.CapMove,
 				imap.CapStatusSize,
 				imap.CapBinary,
+				imap.CapCondStore,
```
emersion, line 78: This list is for extensions only suitable for IMAP4rev1, ie. extensions which don't make sense with IMAP4rev2 (e.g. because they have been folded into IMAP4rev2).

`CONDSTORE` makes sense for IMAP4rev2, so I think it should be move below, outside the `if available.Has(imap.CapIMAP4rev1)`.

imapserver/fetch.go
```diff
@@ -75,6 +75,21 @@ func (c *Conn) handleFetch(dec *imapwire.Decoder, numKind NumKind) error {
...
 	}
 
+	if dec.SP() && dec.Special('(') {
+		var param string
+		if !dec.ExpectAtom(&param) {
+			return dec.Err()
+		}
+
+		if strings.ToUpper(param) == "CHANGEDSINCE" {
+			if !dec.ExpectSP() || !dec.ExpectModSeq(&options.ChangedSince) || !dec.ExpectSpecial(')') {
```
emersion, line 85: Ditto: can `if !dec.ExpectSpecial(')')` be moved below, out of `if strings.ToUpper(param) == "CHANGEDSINCE"`?

imapserver/fetch.go
```diff
@@ -75,6 +75,21 @@ func (c *Conn) handleFetch(dec *imapwire.Decoder, numKind NumKind) error {
...
+				return dec.Err()
+			}
```
emersion, line 87: To ease server implementations, could we set `options.ModSeq = true` here as well? From the RFC:

> When the CHANGEDSINCE FETCH modifier is specified, it implicitly
      adds the MODSEQ FETCH message data item

imapserver/search.go
```diff
@@ -328,6 +328,37 @@ func readSearchKeyWithAtom(criteria *imap.SearchCriteria, dec *imapwire.Decoder,
...
+				return dec.Err()
+			}
+			metadataType = imap.SearchCriteriaMetadataType(strings.ToLower(typeName))
+			if !dec.ExpectSP() {
+				return dec.Err()
+			}
+		}
+
+		var modSeq int64
+		if !dec.ExpectNumber64(&modSeq) {
```
emersion, line 353: Any reason to not use `ExpectModSeq` here?

imapserver/search.go
```diff
@@ -328,6 +328,37 @@ func readSearchKeyWithAtom(criteria *imap.SearchCriteria, dec *imapwire.Decoder,
 		criteria.Or = append(criteria.Or, or)
 	case "$":
 		criteria.UID = append(criteria.UID, imap.SearchRes())
+	case "MODSEQ":
+		if !dec.ExpectSP() {
+			return dec.Err()
+		}
+		var quotedName, name string
```
emersion, line 335: I don't think we need two variables for the name here? A single should be enough?

If `Quoted` doesn't find a quoted string, it leaves the pointer unchanged.

imapserver/search.go
```diff
@@ -328,6 +328,37 @@ func readSearchKeyWithAtom(criteria *imap.SearchCriteria, dec *imapwire.Decoder,
...
+		var metadataType imap.SearchCriteriaMetadataType
+		if dec.Quoted(&quotedName) {
+			name = quotedName
+			if !dec.ExpectSP() {
+				return dec.Err()
+			}
+			var typeName string
+			if !dec.ExpectAtom(&typeName) {
```
emersion, line 343: Nit: this can be combined with `ExpectSP` above.

imapserver/status.go
```diff
@@ -86,6 +86,9 @@ func (c *Conn) writeStatus(data *imap.StatusData, options *imap.StatusOptions) e
 	if options.NumRecent {
 		listEnc.Item().Atom("RECENT").SP().Number(*data.NumRecent)
 	}
+	if options.HighestModSeq {
+		listEnc.Item().Atom("HIGHESTMODSEQ").SP().Number64(int64(data.HighestModSeq))
```
emersion, line 90: We should be able to use `ModSeq` instead of `Number64` here. This would avoid sending a negative number if a modseq is high.

emersion (commented):
Thanks for the patch! I've reviewed the imapserver bits, I'll read the imapmemserver part a bit later.

I'll also try running imaptest against this PR.

PR #690, opened 2025-06-06
imapserver: Add CONDSTORE and QRESYNC extension support

dejanstrbac:
Implementing changes requested


emersion:
Ref https://github.com/emersion/go-imap/pull/687 which contained the previous version of this PR.

imapserver/select.go
```diff
@@ -69,6 +96,19 @@ func (c *Conn) handleSelect(tag string, dec *imapwire.Decoder, readOnly bool) er
 		}
 	}
 
+	shouldSendModSeqStatus := c.enabled.Has(imap.CapIMAP4rev2) || c.server.options.caps().Has(imap.CapCondStore)
```
emersion: I don't think IMAP4rev2 has CONDSTORE folded in? Is there a reason to send mod seq when it's enabled?

imapserver/imapmemserver/mailbox.go
```diff
@@ -156,6 +161,9 @@ func (mbox *Mailbox) appendBytes(buf []byte, options *imap.AppendOptions) *imap.
 	msg.uid = mbox.uidNext
 	mbox.uidNext++
 
+	msg.modSeq = mbox.highestModSeq
+	mbox.highestModSeq++
```
emersion: Shouldn't we bump `mbox.highestModSeq` before setting the message's mod seq?

If we bump `mbox.highestModSeq` after, then it's not the highest mod seq, it's the next mod seq.

imapserver/imapmemserver/mailbox.go
```diff
@@ -324,6 +333,10 @@ func (mbox *MailboxView) Fetch(w *imapserver.FetchWriter, numSet imap.NumSet, op
 			return
 		}
 
+		if options.ChangedSince > 0 && msg.modSeq <= options.ChangedSince {
```
emersion, line 372: Shouldn't this be `msg.modSeq < options.ChangedSince`? In other words, we want to send the message if CHANGEDSINCE equals MODSEQ.

imapserver/imapmemserver/mailbox.go
```diff
@@ -418,13 +431,40 @@ func (mbox *MailboxView) staticSearchCriteria(criteria *imap.SearchCriteria) {
 }
 
 func (mbox *MailboxView) Store(w *imapserver.FetchWriter, numSet imap.NumSet, flags *imap.StoreFlags, options *imap.StoreOptions) error {
-	mbox.forEach(numSet, func(seqNum uint32, msg *message) {
-		msg.store(flags)
-		mbox.Mailbox.tracker.QueueMessageFlags(seqNum, msg.uid, msg.flagList(), mbox.tracker)
+	var modifiedSeqNums []uint32
+	var modifiedMsgs []*message
+
+	mbox.mutex.Lock()
```
emersion: Nit: we can use `mbox.forEach` instead of manually locking/unlocking.

imapserver/imapmemserver/mailbox.go
```diff
@@ -418,13 +431,40 @@ func (mbox *MailboxView) staticSearchCriteria(criteria *imap.SearchCriteria) {
...
+	mbox.forEachLocked(numSet, func(seqNum uint32, msg *message) {
+		if options != nil && options.UnchangedSince > 0 && msg.modSeq > options.UnchangedSince {
+			return
+		}
+
+		if changed := msg.store(flags); changed {
+			mbox.highestModSeq++
+			msg.modSeq = mbox.highestModSeq
```
emersion: Could we handle the mod seq bump inside `msg.store`? This would make it so we don't forget bumping it if we call this function from elsewhere.

imapserver/imapmemserver/mailbox.go
```diff
@@ -418,13 +431,40 @@ func (mbox *MailboxView) staticSearchCriteria(criteria *imap.SearchCriteria) {
...
+			modifiedMsgs = append(modifiedMsgs, msg)
+		}
 	})
-	if !flags.Silent {
-		return mbox.Fetch(w, numSet, &imap.FetchOptions{Flags: true})
+	mbox.mutex.Unlock()
+
+	for i, seqNum := range modifiedSeqNums {
+		msg := modifiedMsgs[i]
+		mbox.Mailbox.tracker.QueueMessageFlags(seqNum, msg.uid, msg.flagList(), mbox.tracker)
```
emersion: Is there a reason to move the `QueueMessageFlags` call outside of the `mbox.forEach` loop?

`msg.flagList` can only be called with the mailbox mutex locked. `QueueMessageFlags` never blocks.

imapserver/imapmemserver/mailbox.go
```diff
@@ -418,13 +431,40 @@ func (mbox *MailboxView) staticSearchCriteria(criteria *imap.SearchCriteria) {
...
 	}
+
+	if !flags.Silent && len(modifiedMsgs) > 0 {
+		for i, seqNum := range modifiedSeqNums {
```
emersion: Instead of open-coding this, could we accumulate an `imap.SeqSet` of modified message sequence numbers in the loop above, and pass that to `mbox.Fetch`?

emersion (commented):
LGTM apart from these comments!

imapserver/imapmemserver/mailbox.go
dejanstrbac: The RFC specifies that the message's modSeq must be strictly greater than the CHANGEDSINCE value (we have the opposite condition)

imapserver/imapmemserver/mailbox.go
dejanstrbac: update the whole logic with an accumulator

imapserver/imapmemserver/mailbox.go
dejanstrbac: agreed, updated

imapserver/imapmemserver/mailbox.go
dejanstrbac: agreed, updated

imapserver/imapmemserver/mailbox.go
dejanstrbac: right, fixed

imapserver/select.go
dejanstrbac: IMAP4rev2 requires QRESYNC which on the other hand requires CONDSTORE. I added support in capability check for this chain

imapserver/select.go
```diff
@@ -100,7 +192,7 @@ func (c *Conn) handleUnselect(dec *imapwire.Decoder, expunge bool) error {
 	}
 
 	if expunge {
-		w := &ExpungeWriter{}
+		w := &ExpungeWriter{conn: c}
```
dejanstrbac, line 195: @emersion this is an unrelated bug, conn was missing

WhyNotHugo:
I notice that this branch produces:

    UID FETCH 1:* (FLAGS) (CHANGEDSINCE 123) VANISHED

I believe the following to be correct:

    UID FETCH 1:* (FLAGS) (CHANGEDSINCE 123 VANISHED)

From: https://datatracker.ietf.org/doc/html/rfc7162#section-3.2.6:

> The VANISHED UID FETCH modifier MUST only be specified together with the CHANGEDSINCE UID FETCH modifier.

WhyNotHugo:
There's also an example in that section of the RFC:

```
   C: s100 UID FETCH 300:500 (FLAGS) (CHANGEDSINCE 12345
       VANISHED)
```

imapserver/select.go
emersion: [RFC 9051 section 6.4.1](https://www.rfc-editor.org/rfc/rfc9051.html#section-6.4.1):

> No untagged EXPUNGE responses are sent.

Maybe a comment would be warranted here.

emersion:
Could we try merging CONDSTORE first, then do QRESYNC as a second step? Both at the same time are a bit much to review, especially since commits aren't split into semantic chunks. (I'd suggest opening a new draft PR based on this one for QRESYNC, then dropping the QRESYNC stuff from this PR. I can help with that if needed.)

aleksandr4842:
@dejanstrbac @mercata @emersion may be need help?

PR #691, opened 2025-06-06
imapserver: add support for special-use

dejanstrbac:


imapserver/imapmemserver/mailbox.go
```diff
@@ -167,6 +174,13 @@ func (mbox *Mailbox) SetSubscribed(subscribed bool) {
 	mbox.mutex.Unlock()
 }
 
+// SetSpecialUse sets the special-use attributes for this mailbox.
+func (mbox *Mailbox) SetSpecialUse(attrs ...imap.MailboxAttr) {
```
emersion: This doesn't map to an existing IMAP command, so I'd prefer to leave it out...

Could we instead populate the list from `CreateOptions.SpecialUse` at mailbox creation time?

imapserver/imapmemserver/user.go
```diff
@@ -56,6 +56,10 @@ func (u *User) mailbox(name string) (*Mailbox, error) {
 	return u.mailboxLocked(name)
 }
 
+func (u *User) Mailbox(name string) (*Mailbox, error) {
```
emersion: With the above change, this should no longer be necessary.

emersion (commented):
Apart from this, LGTM!

imapserver/imapmemserver/user.go
dejanstrbac: agreed, updated!

emersion (approved):
Thanks!

PR #692, opened 2025-06-06
imapserver: advertise APPENDLIMIT capability

dejanstrbac:


cmd/imapmemserver/main.go
```diff
@@ -68,8 +68,9 @@ func main() {
 			return memServer.NewSession(), nil, nil
 		},
 		Caps: imap.CapSet{
-			imap.CapIMAP4rev1: {},
-			imap.CapIMAP4rev2: {},
+			imap.CapIMAP4rev1:                 {},
+			imap.CapIMAP4rev2:                 {},
+			imap.Cap("APPENDLIMIT=104857600"): {},
```
emersion: I'd prefer to add a new `Session` interface instead of making the servers responsible for crafting this string. This would avoid hardcoding strings and would allow the server to customize the limit depending on the connection.

```go
// SessionAppendLimit is an IMAP session which has the same APPEND limit for
// all mailboxes.
type SessionAppendLimit interface {
	Session
	AppendLimit() uint32
}
```

emersion (commented):
It seems like this implements only part of `APPENDLIMIT`: this is missing the STATUS extension.

cmd/imapmemserver/main.go
dejanstrbac: updated to use SessionAppendLimit along with option to disclose/hide limit

imapserver/session.go
```diff
@@ -114,3 +114,18 @@ type SessionUnauthenticate interface {
...
+	Session
+
+	// AppendLimit returns the maximum size in bytes that can be uploaded to this server
+	// in an APPEND command.
+	AppendLimit() uint32
+
+	// DiscloseLimit indicates whether the limit should be advertised in the CAPABILITY
+	// response. If false, only "APPENDLIMIT" will be listed, without the actual limit.
+	// If true, "APPENDLIMIT=<limit>" will be listed.
+	DiscloseLimit() bool
```
emersion, line 130: What's the purpose of this method? Servers can already reject literals which are too large in APPEND without advertising a limit (by returning an error).

imapserver/status.go
```diff
@@ -45,6 +45,14 @@ func (c *Conn) handleStatus(dec *imapwire.Decoder) error {
 		return err
 	}
 
+	// Check if AppendLimit was requested and the session supports it
+	if options.AppendLimit && data.AppendLimit == nil {
+		if appendLimitSession, ok := c.session.(SessionAppendLimit); ok {
+			limit := appendLimitSession.AppendLimit()
+			data.AppendLimit = &limit
+		}
+	}
```
emersion: This prevents backends from returning a per-mailbox append limit.

Backends should be able to populate `data.AppendLimit` on their own already, so I don't think we need this?

imapserver/session.go
dejanstrbac: APPENDLIMIT capability can be advertised as APPENDLIMIT only or with explicit limit APPENDLIMIT=<limit>. In first you have to try to discover the limit while in the second you know before you try. The method allows to server to configure whether it will advertise it or not.

imapserver/status.go
dejanstrbac: per-mailbox limit is a very edge functionality. While it is possible per RFC, I would not follow it blindly. Dovecot has this as global for instance too. 

Nevertheless, agreed, lets move that to implementing server

dejanstrbac: (updated)

emersion: > per-mailbox limit is a very edge functionality. While it is possible per RFC, I would not follow it blindly. Dovecot has this as global for instance too.

I can agree with that. If it helps, I'd be totally fine with leaving per-mailbox append limit unimplemented until someone needs it.

imapserver/session.go
emersion: Hm, I'm not sure I'm following. If APPENDLIMIT is listed in `Caps`, then we already know that the server wants to disclose the limit?

How about the following:

- If `AppendLimit` is not listed in `Caps`, never advertise `AppendLimit`
- If `AppendLimit` is listed in `Caps`:
  - If `SessionAppendLimit` is implemented and returns a non-zero value, advertise `APPENDLIMIT=n`
  - Otherwise, advertise `APPENDLIMIT` without a value

imapserver/capability.go
```diff
@@ -86,6 +88,21 @@ func (c *Conn) availableCaps() []imap.Cap {
...
 		// Capabilities which require backend support and apply to both
 		// IMAP4rev1 and IMAP4rev2
+		if appendLimitSession, ok := c.session.(SessionAppendLimit); ok {
+			if appendLimitSession.DiscloseLimit() {
+				limit := appendLimitSession.AppendLimit()
+				caps = append(caps, imap.Cap(fmt.Sprintf("APPENDLIMIT=%d", limit)))
+			} else {
+				caps = append(caps, imap.CapAppendLimit)
+			}
+		} else if limit, ok := available.AppendLimit(); ok {
```
emersion, line 98: This adds two ways to set the global limit, either with a `SessionAppendLimit`, either with manual string formatting in `Caps`. I'd prefer to keep only the former.

imapserver/capability.go
```diff
@@ -86,6 +88,21 @@ func (c *Conn) availableCaps() []imap.Cap {
 
 		// Capabilities which require backend support and apply to both
 		// IMAP4rev1 and IMAP4rev2
```
emersion, line 90: Can we keep this comment next to the `addAvailableCaps` list below?

imapserver/capability.go
```diff
@@ -86,6 +88,21 @@ func (c *Conn) availableCaps() []imap.Cap {
...
+		if appendLimitSession, ok := c.session.(SessionAppendLimit); ok {
```
emersion, line 91: Per the RFC, we should only advertise a limit in the authenticated state.

imapserver/imapmemserver/session.go
```diff
@@ -17,13 +17,35 @@ type (
...
-	return &UserSession{user: user}
+	return &UserSession{
+		user:          user,
+		appendLimit:   104857600, // 100 MiB default
+		discloseLimit: true,      // By default, disclose the limit in CAPABILITY
+	}
+}
+
+// NewUserSessionWithAppendLimit creates a new user session with a custom append limit.
+func NewUserSessionWithAppendLimit(user *User, appendLimit uint32, discloseLimit bool) *UserSession {
```
emersion, line 43: Could we keep this simple and only have a single global append limit in imapmemserver?

emersion:
We should also replace this constant: https://github.com/emersion/go-imap/blob/60b13facc937efa83a8101db2446c786d3944fa4/imapserver/append.go#L16

emersion:
I've fixed up the remaining issues and pushed this as 1a15e079f48ba33cdeefe39db409cda3d1d5d1a2.

Next time, please tick the "allow edits from maintainers" checkbox when opening a PR!

PR #693, opened 2025-06-06
imapserver: add ID capability

dejanstrbac:
These simple capabilities are not depending on go-imap and can be made available?

### DIFF
--- a/imapserver/capability.go
+++ b/imapserver/capability.go
@@ -80,7 +80,6 @@ func (c *Conn) availableCaps() []imap.Cap {
 				imap.CapMove,
 				imap.CapStatusSize,
 				imap.CapBinary,
-				imap.CapChildren,
 				imap.CapID,
 			})
 		}
--- a/imapserver/capability.go
+++ b/imapserver/capability.go
@@ -75,6 +75,8 @@ func (c *Conn) availableCaps() []imap.Cap {
 				imap.CapMove,
 				imap.CapStatusSize,
 				imap.CapBinary,
+				imap.CapChildren,
+				imap.CapID,
 			})
 		}
 		addAvailableCaps(&caps, available, []imap.Cap{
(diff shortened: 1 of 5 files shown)


imapserver/capability.go
```diff
@@ -75,6 +75,8 @@ func (c *Conn) availableCaps() []imap.Cap {
 				imap.CapMove,
 				imap.CapStatusSize,
 				imap.CapBinary,
+				imap.CapChildren,
```
emersion: Could CHILDREN be submitted independently in a separate PR? That way, it can be merged right away.

imapserver/capability.go
```diff
@@ -75,6 +75,8 @@ func (c *Conn) availableCaps() []imap.Cap {
...
+				imap.CapID,
```
emersion: The ID extension is applicable to IMAP4rev2 as well, so should be moved below.

I've added a commit to better explain what these cap lists are: b33055e420b742b890c8a93e9a5b50163845bca3

emersion:
What is the motivation for adding ID?

I find this extension to be harmful in general, because it's privacy-intrusive and enables both sides to write workarounds instead of proper fixes for bugs and non-standard behavior.

dejanstrbac:
> What is the motivation for adding ID?
> 
> I find this extension to be harmful in general, because it's privacy-intrusive and enables both sides to write workarounds instead of proper fixes for bugs and non-standard behavior.

It's for debugging and client support only. Often they have multiple clients connecting so knowing which client they use for which connection helps trace their possible issues. It's also helpful to know which IP / which client so they can have a simple security audit. I don't think the worry for privacy belongs here - the trust is assumed on IMAP service between client and server by definition.

imapserver/capability.go
dejanstrbac: moved children to own PR

dejanstrbac:
I've updated the PR for Dovecot forwarding support which is more critical use of ID as per 
[Dovecot Docs](https://doc.dovecot.org/2.3/configuration_manual/forwarding_parameters/)

PR #696, opened 2025-06-09
imapserver: include CHILDREN capability

dejanstrbac:


imapserver/capability.go
```diff
@@ -81,6 +81,7 @@ func (c *Conn) availableCaps() []imap.Cap {
 			imap.CapCreateSpecialUse,
 			imap.CapLiteralPlus,
 			imap.CapUnauthenticate,
+			imap.CapChildren,
```
emersion: In #693, CHILDREN was in the list of caps that only make sense with IMAP4rev1, and that don't make sense in IMAP4rev2. This was correct AFAIU, because IMAP4rev2 also defines these attributes and additionally a CHILDREN list return option.

dejanstrbac: After #693 I've checked again in RFC if it should be folder into rev2; while it does use \HasChildren and \HasNoChildren rev2 does not advertise CHILDREN capability and you have to explicitly advertise its support. Hence better to have it explicitly here for both rev1 and rev2

emersion: My reading of the RFCs is:

- The CHILDREN capability indicates that the server may send the `\HasChildren` and `\HasNoChildren` mailbox attributes
- The IMAP4rev2 capability indicates that the server may send the `\HasChildren` and `\HasNoChildren` mailbox attributes and also supports the LIST CHILDREN return option

So I don't think it makes sense to advertise CHILDREN together with IMAP4rev2?

dejanstrbac: ok, added to rev1 only and folded in rev2 caps.

emersion (approved):
Thank you!

PR #698, opened 2025-07-08
v2: imapclient: add option to specify a custom dialer

fho:
The default 30s dial timeout is too short for slow IMAP servers. When I connect to my provider's server, I often have to retry multiple times until no context deadline error happens.

Make dialer options, like the dial timeout, configurable by adding a Dialer field to the Options struct.
If Dialer is nil, the default dialer with a timeout of 30s is used (as before).

### DIFF
--- a/imapclient/client.go
+++ b/imapclient/client.go
@@ -45,11 +45,9 @@ const (
 
 	cmdWriteTimeout     = 30 * time.Second
 	literalWriteTimeout = 5 * time.Minute
-)
 
-var dialer = &net.Dialer{
-	Timeout: 30 * time.Second,
-}
+	defaultDialTimeout = 30 * time.Second
+)
 
 // SelectedMailbox contains metadata for the currently selected mailbox.
 type SelectedMailbox struct {
@@ -77,6 +75,9 @@ type Options struct {
 	UnilateralDataHandler *UnilateralDataHandler
 	// Decoder for RFC 2047 words.
 	WordDecoder *mime.WordDecoder
+	// Dialer to use when establishing connections with the Dial* functions.
+	// If nil, a default dialer with a 30 second timeout is used.
+	Dialer *net.Dialer
 }
 
 func (options *Options) wrapReadWriter(rw io.ReadWriter) io.ReadWriter {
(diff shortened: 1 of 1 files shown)


imapclient/client.go
```diff
@@ -45,11 +45,9 @@ const (
 
 	cmdWriteTimeout     = 30 * time.Second
 	literalWriteTimeout = 5 * time.Minute
-)
 
-var dialer = &net.Dialer{
-	Timeout: 30 * time.Second,
-}
+	defaultDialTimeout = 30 * time.Second
```
fho, line 49: I prefer to make `defaultDialTimeout` public, wdyt?

emersion:
- `New` and `NewStartTLS` can be used to create a client for an existing `net.Conn`.
- If we do want to make this customizable, probably it would make more sense to take a `net.Dialer` instead of just the timeout.

imapclient/client.go
emersion: Why?

fho:
> * `New` and `NewStartTLS` can be used to create a client for an existing `net.Conn`.
> 
>     * If we do want to make this customizable, probably it would make more sense to take a `net.Dialer` instead of just the timeout.

Yes, agree!
I would then replace the `conn net.Conn` parameter of `New` and `NewStartTLS` with a `*net.Dialer`, ok?

emersion:
> I would then replace the conn net.Conn parameter of New and NewStartTLS with a *net.Dialer, ok?

I don't think that makes sense. `New` and `NewStartTLS` are more flexible than providing a `net.Dialer`. One can dial themselves and pass the resulting `net.Conn` to `New`/`NewStartTLS`, or use something else entirely to build the `net.Conn`.

fho:
> > I would then replace the conn net.Conn parameter of New and NewStartTLS with a *net.Dialer, ok?
> 
> I don't think that makes sense. `New` and `NewStartTLS` are more flexible than providing a `net.Dialer`. One can dial themselves and pass the resulting `net.Conn` to `New`/`NewStartTLS`, or use something else entirely to build the `net.Conn`.

So you suggestion is to add to the `Options` struct a `*net.Dialer` field instead of `DialTimeout`?

emersion:
Yes!

imapclient/client.go
fho: - allows to use the upstream defaultDialTimeout and keeping it in-sync when passing a custom net.Dialer to only change another option than the timeout,
- allows to base other timeouts on it while using the default without having to pass custom Options,
- allows to log the timeout without having to set it explicitly to ensure it stays in-sync

emersion: I think I'd still prefer to keep it private, because hardcoding it to 30s in clients that care makes sense to me, and there are other aspects of `net.Dialer` that we might switch up and wouldn't be covered.

emersion:
Thanks!

PR #699, opened 2025-07-08
ci: adapt dovecot config file to dovecot 2.4

fho:
In alpine/latest dovecot was updated to version 2.4. The configuration file
syntax changes[^1] made the dovecot.conf, used by tests, invalid.

- The configuration file is adapted to the new syntax.
- The mailbox separator used in TestACL is changed to "/". "." Dovecot is
  configured to use "." as namespace separator. It's not clear why it worked
  before, maybe Dovecot was more lenient.

- TestACL now terminates immediately when the SetACL or GetACL commands fail
  instead of continuing the test.

Adapt the configuration file to the 2.4 syntax.

[^1]: https://doc.dovecot.org/main/installation/upgrade/2.3-to-2.4.html

### DIFF
--- a/imapclient/acl_test.go
+++ b/imapclient/acl_test.go
@@ -82,13 +82,13 @@ func TestACL(t *testing.T) {
 			// execute SETACL command
 			err := client.SetACL(tc.mailbox, testUsername, tc.setRightsModification, tc.setRights).Wait()
 			if err != nil {
-				t.Errorf("SetACL().Wait() error: %v", err)
+				t.Fatalf("SetACL().Wait() error: %v", err)
 			}
 
 			// execute GETACL command to reset cache on server
 			getACLData, err := client.GetACL(tc.mailbox).Wait()
 			if err != nil {
-				t.Errorf("GetACL().Wait() error: %v", err)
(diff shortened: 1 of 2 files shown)


emersion:
Should this be before the log path?

fho:
@emersion you are fast :-D 
yes!

emersion:
Unfortunately it seems like this is not enough to make Dovecot happy…

fho:
ACL tests are now failing with:
```
acl_test.go:85: SetACL().Wait() error: imap: NO [NONEXISTENT] Mailbox doesn't exist: MyFolder.Child (0.001 + 0.000 secs).
```

ACLs should be enabled though, I would also expect another error message if ACL support is missing

fho:
When creating the mailbox we use `/` as separator but in the imap commands `.` is used.
That seems suspicious.
Dovecot should be using `.` as separator[^1] except for INBOX where `/` is configured.


[^1]: https://doc.dovecot.org/2.3/configuration_manual/namespace/#core_setting-namespace/separator

emersion (approved):
Thanks!

PR #700, opened 2025-07-16
v2: imapserver: add custom logout

sashamelentyev:
Sometimes we need send request to auth system about session is over. For example send token to auth system about user is log out. But in current lib version we can't do it


emersion:
Hm, `Close` is already called when the session is closed. The `LOGOUT` command is one way to close the connection, it doesn't just logs out (there is `Unauthenticate` for this purpose). The connection might get closed some other way (e.g. timeout).

Is there a reason why `Close` doesn't work for you?

sashamelentyev:
Unauth is way for me. Thank you!

PR #702, opened 2025-07-24
[server] list utf7 decoding is not needed and will panic on tests

dejanstrbac:
- failed: Expected tagged reply 'ok', got 'NO [SERVERBUG] Internal server error'
 - Command (tag 5.9): list "" !#$$&'+,-.0123456789:;<=>?@^_`|[}%*

### DIFF
--- a/imapserver/list.go
+++ b/imapserver/list.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/emersion/go-imap/v2"
 	"github.com/emersion/go-imap/v2/internal/imapwire"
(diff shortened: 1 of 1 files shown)


emersion:
Hm, I'm not sure UTF-7 decoding isn't needed. What if a client wants to list all mailboxes with the pattern "Caffè *"?

dejanstrbac:
seems this is a much wider issue - we should probably check if UTF8 is enabled for the session and if not only then fallback to UTF7 encoding, otherwise we can return as is?

PR #704, opened 2025-08-02
imapclient: fix parsing of BODYSTRUCTURE responses from Dovecot for message/global

andybalholm:
BODYSTRUCTURE responses from Dovecot for message/global attachments don't include the extra data that would be included for message/rfc822 attachments. So parsing them was returning an error. This PR makes that data optional.

### DIFF
--- a/imapclient/fetch.go
+++ b/imapclient/fetch.go
@@ -990,7 +990,11 @@ func readBodyType1part(dec *imapwire.Decoder, typ string, options *Options) (*im
 		return &bs, nil
 	}
 
-	if strings.EqualFold(bs.Type, "message") && (strings.EqualFold(bs.Subtype, "rfc822") || strings.EqualFold(bs.Subtype, "global")) {
+	// If the Content-Type is message/rfc822, the envelope, body structure, and number of
+	// lines must come next. If it's message/global, IMAP4rev2 servers will include them,
+	// but IMAP4rev1 servers may not. So in that case, we can go by whether the next byte
+	// is '(', since the envelope is a parenthesized list.
+	if strings.EqualFold(bs.Type, "message") && (strings.EqualFold(bs.Subtype, "rfc822") || strings.EqualFold(bs.Subtype, "global") && dec.NextByteIs('(')) {
 		var msg imap.BodyStructureMessageRFC822
 
 		msg.Envelope, err = readEnvelope(dec, options)
--- a/imapclient/parse_test.go
+++ b/imapclient/parse_test.go
@@ -1,0 +1,211 @@
+package imapclient
+
+import (
+	"bufio"
+	"encoding/json"
+	"reflect"
+	"strings"
+	"testing"
+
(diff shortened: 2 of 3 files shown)


emersion:
Marking as non-standard because this contradicts the spec:

```
body-type-msg   = media-message SP body-fields SP envelope
                  SP body SP body-fld-lines
media-message   = DQUOTE "MESSAGE" DQUOTE SP
                  DQUOTE ("RFC822" / "GLOBAL") DQUOTE
                    ; Defined in [MIME-IMT]
```

Would be nice to report a bug to Dovecot. Also we should add a comment to explain why we deviate from the spec.

imapclient/fetch.go
```diff
@@ -994,25 +994,26 @@ func readBodyType1part(dec *imapwire.Decoder, typ string, options *Options) (*im
 		var msg imap.BodyStructureMessageRFC822
 
 		msg.Envelope, err = readEnvelope(dec, options)
-		if err != nil {
-			return nil, err
-		}
+		if err == nil {
```
emersion: `err` is ignored when non-nil. We should never throw away an error value: it can happen for many reasons, e.g. I/O error.

Instead, we should only parse the envelope if there's one.

andybalholm: How can we detect whether the envelope is there or not? All I could figure out to do was to try to parse it and see if I got an error.

andybalholm:
According to [section 6 of RFC 9755](https://datatracker.ietf.org/doc/html/rfc9755#name-fetch-bodystructure-and-mes), treating message/global as equivalent to message/rfc822 is only required for IMAP4rev2. It's not quite clear to me, but it might not even be *allowed* under rev1 unless UTF8=ACCEPT has been enabled.

imapclient/fetch.go
andybalholm: Does imapwire.Decoder have any sort of Peek or Unread feature?

andybalholm:
Adding a method to imapwire.Decoder enabled a much cleaner implementation.

andybalholm:
I disagree with the "non-standard" tag on this PR. In RFC 3501 (IMAP4rev1), the definition of `media-message` is different:

```
media-message   = DQUOTE "MESSAGE" DQUOTE SP DQUOTE "RFC822" DQUOTE
                    ; Defined in [MIME-IMT]
```

So this change is required for interoperability with IMAP4rev1 servers.

emersion:
Any reason why this has been closed?

andybalholm:
Just that it seemed like you weren't interested in the change, so I thought I might as well close it. I'd like to see it happen, but I don't want to keep pushing for it if you aren't interested.

andybalholm:
I should have discussed the issue more before opening a pull request. I've opened an issue now: https://github.com/emersion/go-imap/issues/741

emersion:
I just haven't had the chance to have another look at this PR. Thanks for opening an issue.

PR #707, opened 2025-08-26
feat: mv append limit to options

sashamelentyev:


sashamelentyev:
@emersion hello!
What you mean about it?

emersion:
I would prefer to add a `Session` method for this, as done in #692. That way, servers can implement per-session limits.

Issue #708, opened 2025-09-08
imapclient should use an interface not net.Dialer


themachine2025:
Currently you can configure the imap client to use a custom net.Dialer but the issue is that you are still stuck with the standard *net.Dialer. For example you cannot use a proxy. I propose to replace the concrete *net.Dialer with an interface

emersion:
You can create a client from an existing `net.Conn`. I'd rather not try to make the convenience functions accommodate for all use-cases.

PR #709, opened 2025-09-08
feature: implement abstract Dialer

themachine2025:
- Introduce Dialer interface to allow custom dialers (e.g., SOCKS5, proxy)
- Add DialTLSContext function to support context-based dialing with timeouts and cancellation
- Refactor DialTLS to call DialTLSContext with context.Background() for backward compatibility
- Prepare IMAP client to support flexible dialers while keeping existing behavior unchanged

### DIFF
--- a/imapclient/client.go
+++ b/imapclient/client.go
@@ -21,6 +21,7 @@ package imapclient
 
 import (
 	"bufio"
+	"context"
 	"crypto/tls"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


emersion:
Sorry, but as I said in https://github.com/emersion/go-imap/issues/708, I don't think this is desirable.

themachine2025:
That's a bummer. I think the my patch didn't stop anyone from using a net.Dialer if someone wants that so it was a net improvement.

PR #711, opened 2025-09-18
Fix fetchLiteralReader deadlock: close channel on any error

### LINKED ISSUE #710: goroutine leak on Fetch
I located a goroutine leak at

https://github.com/emersion/go-imap/blob/90f04ed2776113f0031aee9ef5878f609ae89694/imapclient/fetch.go#L820

with pprof:

```
goroutine 105958 [chan receive, 126 minutes]:
github.com/emersion/go-imap/v2/imapclient.(*Client).handleFetch.func2()
	/go/pkg/mod/.../imapclient/fetch.go:820 +0xc19

```

This leak happens occasionally, I am not able to reproduce  or tell how the server behaves.

The `done` chan is holded by 
`*fetchLiteralReader.ch` which is closed only when `io.EOF` error.

https://github.com/emersion/go-imap/blob/90f04ed2776113f0031aee9ef5878f609ae89694/imapclient/fetch.go#L1319-L1326

The AI copilot tells me that if `*fetchLiteralReader` receives other errors other than an `io.EOF` error, the done is never closed, is he right?

qjebbs:
This is expected to fix #710

### DIFF
--- a/imapclient/fetch.go
+++ b/imapclient/fetch.go
@@ -1318,7 +1318,7 @@ type fetchLiteralReader struct {
 
 func (lit *fetchLiteralReader) Read(b []byte) (int, error) {
 	n, err := lit.LiteralReader.Read(b)
-	if err == io.EOF && lit.ch != nil {
+	if err != nil && lit.ch != nil {
 		close(lit.ch)
 		lit.ch = nil
 	}


emersion:
By doing this we loose the error.

qjebbs:
I don't get it, the assertion affects only the timing of closing the channel, the error is always returned at last.

emersion:
I was worried the error might not be bubbled up to the library user, but you're right, they should get it when reading the literal.

emersion (approved):
Thanks!

PR #712, opened 2025-09-21
imapserver: search fixes

dejanstrbac:
- According to RFC 4731, a server must ignore any unrecognized RETURN options
- handle empty NumSets
- readSearchKeyWithAtom should return errors instead of swallowing them
- tag in search response should be quoted, would cause iOS Mail mishandling
- replaced deprecated strings.Title

### DIFF
--- a/imapclient/search.go
+++ b/imapclient/search.go
@@ -328,8 +328,16 @@ func readESearchResponse(dec *imapwire.Decoder) (tag string, data *imap.SearchDa
 			if isUID {
 				numKind = imapwire.NumKindUID
 			}
-			if !dec.ExpectNumSet(numKind, &data.All) {
-				return "", nil, dec.Err()
+			if dec.SP() {
+				if !dec.ExpectNumSet(numKind, &data.All) {
+					return "", nil, dec.Err()
+				}
+			} else {
+				if isUID {
+					data.All = imap.UIDSet{}
+				} else {
+					data.All = imap.SeqSet{}
+				}
 			}
 			if data.All.Dynamic() {
 				return "", nil, fmt.Errorf("imapclient: server returned a dynamic ALL number set in SEARCH response")
--- a/imapserver/search.go
+++ b/imapserver/search.go
@@ -87,7 +87,22 @@ func (c *Conn) handleSearch(tag string, dec *imapwire.Decoder, numKind NumKind) 
 		return err
 	}
 
-	if c.enabled.Has(imap.CapIMAP4rev2) || extended {
+	var supportsESEARCH bool
+	if capSession, ok := c.session.(SessionCapabilities); ok {
+		sessionCaps := capSession.GetCapabilities()
+		supportsESEARCH = sessionCaps.Has(imap.CapESearch) || sessionCaps.Has(imap.CapIMAP4rev2)
+	} else {
+		availableCaps := c.availableCaps()
+		for _, cap := range availableCaps {
+			if cap == imap.CapESearch || cap == imap.CapIMAP4rev2 {
+				supportsESEARCH = true
+				break
+			}
+		}
+	}
+
+	// Use ESEARCH format only if session supports it AND client used extended syntax
+	if supportsESEARCH && extended {
 		return c.writeESearch(tag, data, &options, numKind)
 	} else {
 		return c.writeSearch(data.All)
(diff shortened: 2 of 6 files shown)


imapserver/search.go
```diff
@@ -178,7 +199,7 @@ func readSearchReturnOpts(dec *imapwire.Decoder, options *imap.SearchOptions) er
 		case "SAVE":
 			options.ReturnSave = true
 		default:
-			return newClientBugError("unknown SEARCH RETURN option")
+			// RFC 4731: A server MUST ignore any unrecognized return options.
```
emersion, line 202: I don't see where this is specified in RFC 4731? This would also be surprising to clients: they ask for something and silently get a reply without it.

RFC 4466 says that return options may be followed by optional `search-mod-params`, so we can't just skip the name anyways.

imapserver/search.go
```diff
@@ -339,5 +368,5 @@ func readSearchKeyWithAtom(criteria *imap.SearchCriteria, dec *imapwire.Decoder,
 }
 
 func searchKeyFlag(key string) imap.Flag {
-	return imap.Flag("\\" + strings.Title(strings.ToLower(key)))
+	return imap.Flag("\\" + cases.Title(language.English).String(strings.ToLower(key)))
```
emersion, line 371: We'll never get anything other than ASCII here. I'd prefer not to introduce a new dependency with large Unicode tables just for this.

imapserver/search.go
```diff
@@ -196,7 +217,15 @@ func readSearchKey(criteria *imap.SearchCriteria, dec *imapwire.Decoder) error {
 		return readSearchKeyWithAtom(criteria, dec, key)
 	}
 	return dec.ExpectList(func() error {
-		return readSearchKey(criteria, dec)
+		for {
+			if err := readSearchKey(criteria, dec); err != nil {
+				return err
+			}
+			if !dec.SP() {
```
emersion, line 224: Why stop on non-spaces?

imapclient/search.go
```diff
@@ -328,8 +328,16 @@ func readESearchResponse(dec *imapwire.Decoder) (tag string, data *imap.SearchDa
 			if isUID {
 				numKind = imapwire.NumKindUID
 			}
-			if !dec.ExpectNumSet(numKind, &data.All) {
-				return "", nil, dec.Err()
+			if dec.SP() {
```
emersion, line 331: Grammar from RFC 4731 says that ALL must always be followed by a seq-set:

```abnf
"ALL" SP sequence-set
```

Is a client not conforming?

imapserver/search.go
```diff
@@ -98,15 +115,15 @@ func (c *Conn) writeESearch(tag string, data *imap.SearchData, options *imap.Sea
 
 	enc.Atom("*").SP().Atom("ESEARCH")
 	if tag != "" {
-		enc.SP().Special('(').Atom("TAG").SP().Atom(tag).Special(')')
+		enc.SP().Special('(').Atom("TAG").SP().String(tag).Special(')')
```
emersion, line 118: Good catch. In IMAP4rev2 this is an `astring`, but in RFC 4466 this must be a `string`.

imapserver/search.go
```diff
@@ -85,7 +87,22 @@ func (c *Conn) handleSearch(tag string, dec *imapwire.Decoder, numKind NumKind)
 		return err
 	}
 
-	if c.enabled.Has(imap.CapIMAP4rev2) || extended {
+	var supportsESEARCH bool
+	if capSession, ok := c.session.(SessionCapabilities); ok {
```
emersion, line 91: Seems like this is the only place where `SessionCapabilities` is used? I would've expected it to affect everything, not just SEARCH.

I do think it would be nice to be able to customize supported caps per-conn, FWIW.

emersion (commented):
This PR mixes multiple unrelated things together. This makes review quite tedious.

go-imap uses [linear, recipe style](https://www.bitsnbites.eu/git-history-work-log-vs-recipe/) commits. Each commit is small and self-contained. Commits start with the name of the package they touch as a prefix.

If a PR does a single thing, I can squash the changes into a single commit and edit the commit message to conform. If the PR does multiple things but they are split into logical commits, I can review commits one-by-one, and merge them incrementally if desired.

emersion:
> - readSearchKeyWithAtom should return errors instead of swallowing them
> - tag in search response should be quoted, would cause iOS Mail mishandling

These fixes LGTM, I've pushed them to v2 as 9139e6b781774428c5189bac71176e91877ded6a and 864c4c3e551d320791c2eba7e3d89d80f862c574!

PR #714, opened 2025-09-21
imapserver: allow empty lines without breaking connection

dejanstrbac:
Currently imapserver will end connection if empty command is given (CR)

### DIFF
--- a/imapserver/conn.go
+++ b/imapserver/conn.go
@@ -194,6 +194,10 @@ func (c *Conn) serve() {
 }
 
 func (c *Conn) readCommand(dec *imapwire.Decoder) error {
+	if dec.CRLF() {
+		return nil // allow empty newlines
+	}
+
 	var tag, name string
 	if !dec.ExpectAtom(&tag) || !dec.ExpectSP() || !dec.ExpectAtom(&name) {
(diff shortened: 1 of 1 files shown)


imapserver/conn.go
```diff
@@ -194,6 +194,17 @@ func (c *Conn) serve() {
 }
 
 func (c *Conn) readCommand(dec *imapwire.Decoder) error {
+	for {
+		if dec.EOF() {
+			return nil
+		}
+
+		if dec.ExpectCRLF() {
```
emersion: The `Expect` family of functions sets the error field in the decoder. Since not getting a CRLF is valid, we should use `Decoder.CRLF` instead.

imapserver/conn.go
```diff
@@ -194,6 +194,17 @@ func (c *Conn) serve() {
 }
 
 func (c *Conn) readCommand(dec *imapwire.Decoder) error {
+	for {
```
emersion: I don't think using a for loop is necessary, plus it won't behave well regarding the command read timeout. We can just return if we got a CRLF and leave it to the caller to invoke `readCommand` again.

emersion:
I've manually resolved the above comments.

PR #718, opened 2025-10-11
imapclient: implement support for NOTIFY

WhyNotHugo:
This patch series implements support for the NOTIFY extension.

The first commit implements support in imapclient, the main focus of my work. I've been using this on a client to monitor changes with success.

The second commit implements support in imapserver.

The third commit adds a minimal scaffold implementation in imapmemserver. This implementation simply rejects any request (which it technically allows as per the spec). It's not an actually _useful_ implementation, and is mostly there so we can have minimal client-server unit tests. Implementing proper NOTIFY support for imapmemserver likely requires substantial changes which fall beyond the scope of this series.

This is likely easiest to review on a per-commit basis, and about half of the LoC are just unit tests for encoding/decoding.

### DIFF
--- a/imapclient/client.go
+++ b/imapclient/client.go
@@ -905,6 +905,10 @@ func (c *Client) readResponseData(typ string) error {
 				}
 			case "NOMODSEQ":
 				// ignore
+			case "NOTIFICATIONOVERFLOW":
+				if handler := c.options.unilateralDataHandler().NotificationOverflow; handler != nil {
+					handler()
+				}
 			default: // [SP 1*<any TEXT-CHAR except "]">]
 				if c.dec.SP() {
 					c.dec.DiscardUntilByte(']')
@@ -1188,14 +1192,29 @@ type UnilateralDataMailbox struct {
 //
 // The handler will be invoked in an arbitrary goroutine.
 //
+// These handlers are important when using the IDLE or NOTIFY commands, as the
+// server will send unsolicited STATUS, FETCH, and EXPUNGE responses for
+// mailbox events.
+//
 // See Options.UnilateralDataHandler.
 type UnilateralDataHandler struct {
 	Expunge func(seqNum uint32)
 	Mailbox func(data *UnilateralDataMailbox)
 	Fetch   func(msg *FetchMessageData)
 
-	// requires ENABLE METADATA or ENABLE SERVER-METADATA
+	// Requires ENABLE METADATA or ENABLE SERVER-METADATA.
 	Metadata func(mailbox string, entries []string)
+
+	// Called when the server sends an unsolicited STATUS response.
+	//
+	// Commonly used with NOTIFY to receive mailbox status updates
+	// for non-selected mailboxes (RFC 5465).
+	Status func(data *imap.StatusData)
+
+	// Called when the server sends NOTIFICATIONOVERFLOW (RFC 5465).
+	//
+	// Indicates the server has disabled all NOTIFY notifications.
+	NotificationOverflow func()
 }
 
 // command is an interface for IMAP commands.
--- a/imapclient/notify.go
+++ b/imapclient/notify.go
@@ -1,0 +1,95 @@
+package imapclient
+
+import (
+	"fmt"
+
+	"github.com/emersion/go-imap/v2"
+	"github.com/emersion/go-imap/v2/internal/imapwire"
+)
+
+// Notify sends a NOTIFY command (RFC 5465).
+//
+// The NOTIFY command allows clients to request server-push notifications
+// for mailbox events like new messages, expunges, flag changes, etc.
+//
+// When NOTIFY SET is active, the server may send unsolicited responses at any
+// time (STATUS, FETCH, EXPUNGE, LIST responses). These unsolicited responses
+// are delivered via the UnilateralDataHandler callbacks set in
+// imapclient.Options.
+//
+// When the server sends NOTIFICATIONOVERFLOW, the NotificationOverflow callback
+// in UnilateralDataHandler will be called (if set).
+func (c *Client) Notify(options *imap.NotifyOptions) (*NotifyCommand, error) {
+	cmd := &NotifyCommand{}
+	enc := c.beginCommand("NOTIFY", cmd)
+	if err := encodeNotifyOptions(enc.Encoder, options); err != nil {
+		enc.end()
+		return nil, err
+	}
+	enc.end()
+
+	return cmd, nil
+}
+
+// encodeNotifyOptions encodes NOTIFY command options to the encoder.
+func encodeNotifyOptions(enc *imapwire.Encoder, options *imap.NotifyOptions) error {
+	if options == nil || len(options.Items) == 0 {
+		// NOTIFY NONE: disable all notifications.
+		enc.SP().Atom("NONE")
+		return nil
+	}
+
+	enc.SP().Atom("SET")
+
+	if options.Status {
+		enc.SP().List(1, func(i int) {
+			enc.Atom("STATUS")
+		})
+	}
+
+	for _, item := range options.Items {
+		if item.MailboxSpec == "" && len(item.Mailboxes) == 0 {
+			return fmt.Errorf("invalid NOTIFY item: must specify either MailboxSpec or Mailboxes")
+		}
+
+		enc.SP().List(1, func(_ int) {
+			if item.MailboxSpec != "" {
+				enc.Atom(string(item.MailboxSpec))
+			} else {
+				// len(item.Mailboxes) > 0, as per the check above.
+				if item.Subtree {
+					enc.Atom("SUBTREE").SP()
+				}
+				enc.List(len(item.Mailboxes), func(j int) {
+					enc.Mailbox(item.Mailboxes[j])
+				})
+			}
+
+			if len(item.Events) > 0 {
+				enc.SP().List(len(item.Events), func(j int) {
+					enc.Atom(string(item.Events[j]))
+				})
+			}
+		})
+
+	}
+
+	return nil
+}
+
+// NotifyCommand is a NOTIFY command.
+//
+// When NOTIFY SET is active, the server may send unsolicited responses at any
+// time. These responses are delivered via UnilateralDataHandler
+// (see Options.UnilateralDataHandler).
+//
+// If the server sends NOTIFICATIONOVERFLOW, the NotificationOverflow callback
+// in UnilateralDataHandler will be called (if set).
+type NotifyCommand struct {
+	commandBase
+}
+
+// Wait blocks until the NOTIFY command has completed.
+func (cmd *NotifyCommand) Wait() error {
+	return cmd.wait()
+}
--- a/imapclient/notify_encode_test.go
+++ b/imapclient/notify_encode_test.go
@@ -1,0 +1,306 @@
+package imapclient
+
+import (
+	"bufio"
+	"bytes"
+	"testing"
+
+	"github.com/emersion/go-imap/v2"
+	"github.com/emersion/go-imap/v2/internal/imapwire"
+)
+
+func encodeToString(options *imap.NotifyOptions) (string, error) {
+	buf := &bytes.Buffer{}
+	bw := bufio.NewWriter(buf)
+	enc := imapwire.NewEncoder(bw, imapwire.ConnSideClient)
+
+	if err := encodeNotifyOptions(enc, options); err != nil {
+		return "", err
+	}
+
+	enc.CRLF()
+	bw.Flush()
+
+	return buf.String(), nil
+}
+
+func TestEncodeNotifyOptions(t *testing.T) {
+	tests := []struct {
+		name     string
+		options  *imap.NotifyOptions
+		expected string
+	}{
+		{
+			name:     "None",
+			options:  nil,
+			expected: " NONE\r\n",
+		},
+		{
+			name: "EmptyItems",
+			options: &imap.NotifyOptions{
+				Items: []imap.NotifyItem{},
+			},
+			expected: " NONE\r\n",
+		},
+		{
+			name: "Selected",
+			options: &imap.NotifyOptions{
+				Items: []imap.NotifyItem{
+					{
+						MailboxSpec: imap.NotifyMailboxSpecSelected,
+						Events: []imap.NotifyEvent{
+							imap.NotifyEventMessageNew,
+							imap.NotifyEventMessageExpunge,
+						},
+					},
+				},
+			},
+			expected: " SET (SELECTED (MessageNew MessageExpunge))\r\n",
+		},
+		{
+			name: "SelectedDelayed",
+			options: &imap.NotifyOptions{
+				Items: []imap.NotifyItem{
+					{
+						MailboxSpec: imap.NotifyMailboxSpecSelectedDelayed,
+						Events: []imap.NotifyEvent{
+							imap.NotifyEventMessageNew,
+							imap.NotifyEventMessageExpunge,
+						},
+					},
+				},
+			},
+			expected: " SET (SELECTED-DELAYED (MessageNew MessageExpunge))\r\n",
+		},
+		{
+			name: "Personal",
+			options: &imap.NotifyOptions{
+				Items: []imap.NotifyItem{
+					{
+						MailboxSpec: imap.NotifyMailboxSpecPersonal,
+						Events: []imap.NotifyEvent{
+							imap.NotifyEventMailboxName,
+							imap.NotifyEventSubscriptionChange,
+						},
+					},
+				},
+			},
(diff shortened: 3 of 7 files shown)


WhyNotHugo:
v2: removed a very long example from the docs. It's very easy for this to fall out of sync and not really useful.

WhyNotHugo:
Tests with dovecot failed. When writing these tests I didn't realise that they also run with another server in CI.

I lean towards dropping the third commit— the one which implements the scaffold for NOTIFY in imapmemserver (along with the test which run server+client).

Thoughts?

WhyNotHugo:
We can also drop the incomplete imapmemserver support and enable the client-server NOTIFY tests only for dovecot. I see the `useDovecot` flag which should be useful for that.

I'll wait for feedback before advancing this this in any way.

WhyNotHugo:
v3: added test for how disconnections are handled (both with and without NOTIFY). Fix `Overflow()` channel hanging on disconnection.

WhyNotHugo:
I've re-written the client test to run only with dovecot, which has real NOTIFY support. Those test will be skipped with the imapmemserver, but are a lot more meaningful than before.

I've ripped out the scaffold for the imapmemserver's NOTIFY support. Such broken support was useless beyond these tests.

notify.go
```diff
@@ -0,0 +1,63 @@
...
+	NotifyEventMessageExpunge   NotifyEvent = "MessageExpunge"
+
+	// Mailbox events
+	NotifyEventMailboxName           NotifyEvent = "MailboxName"
+	NotifyEventSubscriptionChange    NotifyEvent = "SubscriptionChange"
+	NotifyEventMailboxMetadataChange NotifyEvent = "MailboxMetadataChange"
+	NotifyEventServerMetadataChange  NotifyEvent = "ServerMetadataChange"
+)
+
+// NotifyMailboxSpec represents a mailbox specifier (rfc5465#section-6) for the NOTIFY command.
```
emersion: godoc will linkify RFC references when they are written in English: "RFC 5465 section 6"

notify.go
```diff
@@ -0,0 +1,63 @@
...
+	NotifyMailboxSpecPersonal        NotifyMailboxSpec = "PERSONAL"
+	NotifyMailboxSpecInboxes         NotifyMailboxSpec = "INBOXES"
+	NotifyMailboxSpecSubscribed      NotifyMailboxSpec = "SUBSCRIBED"
+)
+
+// NotifyOptions contains options for the NOTIFY command.
+type NotifyOptions struct {
+	// STATUS indicates that a STATUS response should be sent for new mailboxes.
+	// Only valid with Personal, Inboxes, or Subscribed mailbox specs.
+	STATUS bool
```
emersion: Style: in Go we generally use TitleCase instead of UPPERCASE, even if the original spec uses UPPERCASE. (See e.g. Get/Post in net/http.)

imapclient/status.go
```diff
@@ -75,6 +75,11 @@ func (c *Client) handleStatus() error {
 		cmd.pendingData.Status = data
 		cmd.mailboxes <- cmd.pendingData
 		cmd.pendingData = nil
+	default:
+		// Unsolicited STATUS response (e.g., from NOTIFY)
+		if handler := c.options.unilateralDataHandler().Status; handler != nil {
+			handler(data)
+		}
```
emersion: Could we put this in a `if cmd == nil` check after `findPendingCmdFunc`, instead of in a default case? (TBH it would be better to panic in the default case.)

imapclient/client.go
```diff
@@ -1179,14 +1184,24 @@ type UnilateralDataMailbox struct {
 //
 // The handler will be invoked in an arbitrary goroutine.
 //
+// These handlers are important when using the NOTIFY command , as the server
+// will send unsolicited STATUS, FETCH, and EXPUNGE responses for mailbox
+// events.
```
emersion: If we add something about NOTIFY, we should probably add something about IDLE as well.

Typo: space before `,`

imapclient/notify.go
```diff
@@ -0,0 +1,154 @@
...
+			})
+		}
+	}
+}
+
+// NotifyNone sends a NOTIFY NONE command to disable all notifications.
+func (c *Client) NotifyNone() error {
+	_, err := c.Notify(nil)
+	return err
+}
```
emersion: Maybe we can leave it up to the user to call `Notify` with `nil` instead of providing a thin wrapper? The user might be interested in waiting for the server to accept the command, too.

imapclient/notify.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	return cmd, nil
+}
+
+// encodeNotifyOptions encodes NOTIFY command options to the encoder.
+func encodeNotifyOptions(enc *imapwire.Encoder, options *imap.NotifyOptions) {
+	if options == nil || len(options.Items) == 0 {
+		// NOTIFY NONE - disable all notifications
+		enc.SP().Atom("NONE")
```
emersion, line 38: Could use an early `return` to avoid indenting too much

imapclient/notify.go
```diff
@@ -0,0 +1,154 @@
...
+				enc.Atom("STATUS")
+			})
+		}
+
+		// Encode each notify item
+		for _, item := range options.Items {
+			// Validate the item before encoding
+			if item.MailboxSpec == "" && len(item.Mailboxes) == 0 {
+				// Skip invalid items - this shouldn't happen with properly constructed NotifyOptions
+				continue
```
emersion: We should probably return an error here?

imapclient/notify.go
```diff
@@ -0,0 +1,154 @@
...
+			}
+
+			enc.SP().List(1, func(i int) {
+				// Encode mailbox specification
+				if item.MailboxSpec != "" {
+					enc.Atom(string(item.MailboxSpec))
+				} else if len(item.Mailboxes) > 0 {
```
emersion: This should never happen per the above. Would be easier to understand that the mailbox spec or list aren't optional if we dropped this `if`.

notify.go
```diff
@@ -0,0 +1,63 @@
+package imap
+
+const (
+	// ResponseCodeNotificationOverflow is returned when the server cannot
+	// handle the requested notifications (RFC 5465).
+	ResponseCodeNotificationOverflow ResponseCode = "NOTIFICATIONOVERFLOW"
```
emersion: No need to expose this in our public API if it's never going to be returned in a public-facing `ResponseCode` field (e.g. `Error`).

imapclient/client.go
```diff
@@ -896,6 +896,11 @@ func (c *Client) readResponseData(typ string) error {
 				}
 			case "NOMODSEQ":
 				// ignore
+			case "NOTIFICATIONOVERFLOW":
+				// Server has disabled NOTIFY due to overflow (RFC 5465 section 5.8)
+				if cmd := findPendingCmdByType[*NotifyCommand](c); cmd != nil {
+					cmd.handleOverflow()
```
emersion: I think I'd prefer a new callback in the unilateral data handler over a goroutine here…

emersion (commented):
Thanks for the patch! Here are a few comments. I've only read the client part for now, not the tests nor server.

WhyNotHugo:
~~I've added another commit (which probably needs to be squashed later) which allows recognising whether a NOTIFY stream ended because of an explicit NOTIFICATIONOVERFLOW or because the connection closed.~~

I missed the review above. Using UnilateralDataHandler is much cleaner!

imapclient/notify.go
WhyNotHugo: I can't find any other method which uses an encoder and returns an error.

Is it fine to leave the command half-written in this case? (this indicates a programmer error anyway).

emersion: For some commands we `panic` when fed an invalid state (e.g. see `GetMetadata`). For some commands we validate upfront and we return a command struct already completed with an error (e.g. see `Enable`). Latter is more complicated and better suited for cases where the failure doesn't indicate a programmer error (`Enable` is probably a bad usage for this).

WhyNotHugo:
All comments addressed; ready for a second round of review.

imapclient/notify.go
WhyNotHugo: Do you prefer for this to panic or return an error?

emersion: I would prefer to not return an `error` so that callers can chain `.Wait()` like other short-running commands. I don't have strong opinions about returning a command struct already completed with an error vs. a panic. Simplest would probably be to panic. (We can change it later if it turns out to be an annoyance.)

imapclient/client.go
```diff
@@ -896,6 +896,10 @@ func (c *Client) readResponseData(typ string) error {
 				}
 			case "NOMODSEQ":
 				// ignore
+			case "NOTIFICATIONOVERFLOW":
+				if c.options.UnilateralDataHandler.NotificationOverflow != nil {
```
emersion: Don't we need to use `unilateralDataHandler()` here to avoid nil pointer deref when `options.UnilateralDataHandler` is nil?

imapclient/status.go
```diff
@@ -68,13 +68,21 @@ func (c *Client) handleStatus() error {
...
+		return nil
+	}
 	switch cmd := cmd.(type) {
 	case *StatusCommand:
 		cmd.data = *data
 	case *ListCommand:
 		cmd.pendingData.Status = data
 		cmd.mailboxes <- cmd.pendingData
 		cmd.pendingData = nil
+	default:
```
emersion: Nit: no need to add a default case here

imapserver/session.go
```diff
@@ -124,3 +124,11 @@ type SessionAppendLimit interface {
 	// this server in an APPEND command.
 	AppendLimit() uint32
 }
+
+// SessionNotify is an IMAP session which supports NOTIFY.
+type SessionNotify interface {
+	Session
+
+	// Authenticated state
+	Notify(w *UpdateWriter, options *imap.NotifyOptions) error
```
emersion: I wonder about the implications of passing an `UpdateWriter` to the backend here. Up until now backends would only use `UpdateWriter` synchronously in handlers. Here the backend would need to use it after `Notify` returns. We have `conn.encMutex` to avoid stepping on our own toes when sending an update, but that still leaves considerations such as sequence number issues with EXPUNGE notifications while another command is running, see [section 5.5](https://www.rfc-editor.org/rfc/rfc9051.html#section-5.5).

Maybe there is a way to expose a safer interface to backends. Maybe we should just leave it up to the backend and document the issue.

imapclient/connection_test.go
emersion, line 1: I feel like the new connection tests are orthogonal to NOTIFY - if so, could these be sent as a separate PR?

Note, we had a similar test before and it ended up being racy: https://github.com/emersion/go-imap/commit/805dce07ee597fd849e30be686124e37c513a430

WhyNotHugo: https://github.com/emersion/go-imap/pull/727

imapclient/client.go
WhyNotHugo: Good catch. I've also re-written this to match the style of other handler calls.

imapserver/session.go
WhyNotHugo: Would you prefer that I drop the `imapserver` support from this PR and focus on the client part first?

WhyNotHugo: FWIW, I only wrote some of the server portions to test the client, but eventually moved on to just testing with dovecot.

emersion: Moving the server parts to a separate PR would allow the client part to be merged first, which sounds like a good idea.

imapclient/notify_encode_test.go
```diff
@@ -0,0 +1,390 @@
...
+import (
+	"bufio"
+	"bytes"
+	"testing"
+
+	"github.com/emersion/go-imap/v2"
+	"github.com/emersion/go-imap/v2/internal/imapwire"
+)
+
+func encodeToString(options *imap.NotifyOptions) (string, error) {
```
emersion, line 12: Using table-driven tests should allow us to cut down the boiler plate here I believe: https://go.dev/wiki/TableDrivenTests

imapclient/notify_encode_test.go
```diff
@@ -0,0 +1,390 @@
...
+					imap.NotifyEventMessageNew,
+				},
+			},
+		},
+	}
+	result, err := encodeToString(options)
+	if err != nil {
+		t.Fatalf("encodeToString() error = %v", err)
+	}
+	expected := " SET ((INBOX \"Foo Bar\" \"Test&-Mailbox\") (MessageNew))\r\n"
```
emersion: Backtick strings can improve readability here

imapclient/notify_test.go
```diff
@@ -0,0 +1,357 @@
...
+	"os"
+	"testing"
+	"time"
+
+	"github.com/emersion/go-imap/v2"
+	"github.com/emersion/go-imap/v2/imapclient"
+)
+
+func TestClient_Notify(t *testing.T) {
+	if os.Getenv("GOIMAP_TEST_DOVECOT") != "1" {
```
emersion: Can we check whether the extension is supported instead of hardcoding "Dovecot == supported"? See e.g. `TestACL`.

imapclient/notify.go
emersion: It seems like this comment has been addressed?

imapclient/notify_test.go
```diff
@@ -0,0 +1,357 @@
...
+			},
+		},
+	}
+
+	cmd, err := client.Notify(options)
+	if err != nil {
+		t.Fatalf("Notify() = %v", err)
+	}
+
+	if cmd == nil {
```
emersion: Shouldn't we wait for the command to complete in all tests?

imapclient/notify_test.go
```diff
@@ -0,0 +1,357 @@
...
+		t.Skip("NOTIFY test requires GOIMAP_TEST_DOVECOT=1")
+	}
+
+	existsCh := make(chan uint32, 1)
+
+	// Clients needs a UnilateralDataHandler configured, so set it up manually.
+	conn, server := newDovecotClientServerPair(t)
+	defer server.Close()
+
+	options := &imapclient.Options{
```
emersion, line 14: Oh, do we need to pass custom `UnilateralDataHandler` to the client, but `newClientServerPair` doesn't support this?

arnt:
Wow, cool, support for 5465!

I find it difficult to assess the test coverage here, perhaps because of the verbosity. @emersion suggests tables, and I actually wrote the examples in the RFC together with unit tests that might be described as tables. That was C++, in Go I think I'd use an array of 5-tuples:
- a function literal to set up notifications
- a string with the expected/correct IMAP command
- a string with a mailbox to select
- a string that is sent by a mock server after the select has completed
- a function literal that asserts that go-imap acts correctly on the mock server string
 
Each example in section 5 corresponds to one of those tests, and that gave full coverage, easily verifiable. (Rereading now, I notice that the interaction between 5267 and 5465 isn't described in this way. Sigh. Sorry. I didn't have 5267 support myself and I suppose I didn't pay attention.)

Second, I like the use of unilateralDataHandler, but feel that it misses something. It can be nil, but handling unilateral data is almost mandatory now. (If gmail sends you an unsolicited `BYE`, that generally just means that a server there is overloaded and you should reconnect and be routed to another.) Nil seems to be a poor default, but I can't suggest a good way to write a general handler in Go… sorry about the lack of clarity here.

WhyNotHugo:
Tests were verbose, but I originally preferred to keep them that way since they're at least straightforward without too many clever abstractions.

TableDrivenTests is much shorter and even simpler to read, so migrated to that format.

imapclient/notify_test.go
WhyNotHugo: Added support for this in a separate commit.

imapclient/notify_test.go
WhyNotHugo: Done

imapclient/notify_test.go
WhyNotHugo: `Client.Notify` calls `cmd.Wait()` internally. But as I re-think the approach, that's probably a mistake and it shouldn't, correct?

emersion: Generally we avoid calling `Wait` inside command methods, because that prevents commands from being pipelined. For instance a client cannot send a NOTIFY + FETCH in a single roundtrip with this PR's design.

WhyNotHugo: Right, yes, definitely, the original approach here was wrong.

Updated.

emersion (approved):
LGTM, thanks!

emersion:
Ah, it seems like tests are failing...

WhyNotHugo:
One test refactor was wrong, condition `err == cmd.Wait()` should have been `err = cmd.Wait(); err != nil`.

Fixed.

PR #719, opened 2025-10-12
 imapclient: initial VANISHED support 

WhyNotHugo:
Builds on top of https://github.com/emersion/go-imap/pull/718


WhyNotHugo:
Actually, rebasing this away from #718 is trivial. See https://github.com/emersion/go-imap/pull/720

PR #720, opened 2025-10-12
imapclient: initial VANISHED support

WhyNotHugo:


WhyNotHugo:
v2: ran `go fmt`

emersion:
This seems to overlap a bit with https://github.com/emersion/go-imap/pull/690.

WhyNotHugo:
I did a review of that branch (I admit I didn't see it before, because I looked for keyword `VANISHED`.

I only notice one difference, it uses:

    UID FETCH 1:* (FLAGS) (CHANGEDSINCE 123) VANISHED

Whereas my branch uses:

    UID FETCH 1:* (FLAGS) (CHANGEDSINCE 123 VANISHED)

I believe my approach to be correct. From: https://datatracker.ietf.org/doc/html/rfc7162#section-3.2.6:

> The VANISHED UID FETCH modifier MUST only be specified together with the CHANGEDSINCE UID FETCH modifier.

I think it's probably best to fix that in https://github.com/emersion/go-imap/pull/690 and drop this branch.

WhyNotHugo:
Any thoughts on merging this one?

https://github.com/emersion/go-imap/pull/690 has some correctness issues, and hasn't seen some activity in months. I think we can move this smaller PR forward, and eventually rebase the other one on top. It also means reviewing to smaller PRs instead on one bit one :)

PR #721, opened 2025-10-16
imapclient: implement Client.Closed

WhyNotHugo:
Implement a Client.Closed() to be used by clients to monitor when a connection is closed.

### DIFF
--- a/imapclient/client.go
+++ b/imapclient/client.go
@@ -385,6 +385,15 @@ func (c *Client) Mailbox() *SelectedMailbox {
 	return c.mailbox
 }
 
+// Closed returns a channel that is closed when the connection is closed.
+//
+// This channel cannot be used to reliably determine whether a connection is healthy. If
+// the underlying connection times out, the channel will be closed eventually, but not
+// immediately. To check whether the connection is healthy, send a command (such as Noop).
+func (c *Client) Closed() <-chan struct{} {
+	return c.decCh
+}
+
 // Close immediately closes the connection.
 func (c *Client) Close() error {
 	c.mutex.Lock()
--- a/imapclient/example_test.go
+++ b/imapclient/example_test.go
@@ -378,3 +378,34 @@ func ExampleClient_Authenticate_oauth() {
 		log.Fatalf("authentication failed: %v", err)
 	}
 }
+
+func ExampleClient_Closed() {
+	c, err := imapclient.DialTLS("mail.example.org:993", nil)
+	if err != nil {
+		log.Fatalf("failed to dial IMAP server: %v", err)
+	}
+
+	selected := false
+
+	go func(c *imapclient.Client) {
+		if err := c.Login("root", "asdf").Wait(); err != nil {
+			log.Fatalf("failed to login: %v", err)
+		}
+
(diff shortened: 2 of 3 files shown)


WhyNotHugo:
I've implemented by particular case in some other way, but this sounds useful to have. I have no strong objections if you'd prefer to close this PR.

emersion:
I was wondering about the use-case for this. I initially thought I needed something similar for alps, but then figured it wouldn't help much:

- The connection might be broken even when the reader goroutine is still alive, when a connection isn't explicitly closed by the remote and times out.
- Even if the reader goroutine could perfectly detect when a connection is broken, checking for the closed channel introduced in this PR before sending a command would be racy.
- A disconnected client only consumes a small amount of memory, its underlying FD is closed.

WhyNotHugo:
> checking for the closed channel introduced in this PR before sending a command would be racy.

The channel is not intended to be checked before sending a message, but to re-connect automatically when the connection terminates (i.e.: in the case of NOTIFY).

WhyNotHugo:
Previously, the Notify.Overflow channel in https://github.com/emersion/go-imap/pull/718 would also be closed when the connection dropped (a convenient side-effect). This allowed me to detect when a connection had been closed.

Having switched NOTIFY to use a UnilateralDataHandler for NOTIFYOVERFLOW events, I once again need this feature.

I don't think I have any other way of detecting when the connection has been closed. I'm only waiting for NOTIFY events, so never send any messages.

emersion:
Hm, right, that's a good point. I think NOTIFY is special here: with IDLE, clients could call `IdleCommand.Wait` to figure out when the connection was broken. Since NOTIFY isn't a long-running command, there's no equivalent.

imapclient/client.go
```diff
@@ -385,6 +385,11 @@ func (c *Client) Mailbox() *SelectedMailbox {
 	return c.mailbox
 }
 
+// Closed returns a channel that is closed when the connection is closed.
```
emersion, line 388: Could we add a bit more docs about valid usage and expectations? Something like:

> This channel cannot be used to reliably figure out whether a connection is healthy. If the underlying connection errors out, the channel will be closed at some point but the reverse isn't true. To check if the connection is healthy, a command (such as Noop) can be sent.

imapclient/example_test.go
```diff
@@ -378,3 +378,25 @@ func ExampleClient_Authenticate_oauth() {
 		log.Fatalf("authentication failed: %v", err)
 	}
 }
+
+func ExampleClient_Closed() {
```
emersion, line 382: This example is racy (the main function doesn't typically wait for other goroutines to finish) and doesn't reflect real-world usage. Could we drop this example for now, and wait for NOTIFY to be implemented to add a Notify example which leverages Closed to figure out when the connection has been closed?

WhyNotHugo: Example redone.

emersion: The example still feels a bit artificial - but let's revisit this with NOTIFY.

emersion (approved):
Thanks!

PR #722, opened 2025-10-25
extra space between lists of multiparts in bodystructure breaks outlook in ios

dejanstrbac:
Following Dovecot behavior with no space.

PR #723, opened 2025-10-25
content encoding should be traditionally lowercase, not uppercase

dejanstrbac:


emersion:
Does this fix a client-side issue, or is this purely esthetic?

The reason I picked upper-case is the IMAP grammar:

```abnf
body-fld-enc    = (DQUOTE ("7BIT" / "8BIT" / "BINARY" / "BASE64"/
                  "QUOTED-PRINTABLE") DQUOTE) / string
                  ; Content-Transfer-Encoding header field value.
                  ; Defaults to "7BIT" (as per RFC 2045)
                  ; if not present in the body part.
```

dejanstrbac:
Outlook clients have issues with uppercases here. I compared output to Dovecot which uses lowercase. With this and #722 - Outlook is satisfied

emersion:
Reworded the commit message and added a comment, merged as 5c47ad1124b1b0481a5d4e25b7f841668a366beb.

PR #724, opened 2025-11-04
Fixed an issue with ios mail app

YannickFuereder:
the ios mail app needs the " or else it throws a parsing error


emersion:
Hm, are you sure this wasn't already addressed in this commit? https://github.com/emersion/go-imap/commit/9139e6b781774428c5189bac71176e91877ded6a

Adding quotes manually like this doesn't handle escaping correctly. `Quoted` (called by `String`) already adds quotes.

YannickFuereder:
shit i didtn see that

could you push a new release with the fix 🥺

PR #733, opened 2025-12-06
Add recent flag

aleksandr4842:


emersion:
This was intentionally left out from the public API, and is also defined at:

https://github.com/emersion/go-imap/blob/cc7b09a2bf60dc7693c3a71a83f810e2c8193cf3/internal/internal.go#L18

aleksandr4842:
In legacy systems, this flag is still used, unfortunately

arnt:
A question from the sidelines — what client now needs/uses `\recent`? I ask because all I've heard about in the past few years is bugs: "My client reconnected and `\recent` changed", "connection pooling breaks `\recent`" etc.

Issue #491, opened 2023-03-17
v2: client status


emersion:
Continuation of https://github.com/emersion/go-imap/pull/490

- [ ] Consider adding separate types for `FetchItemData{Body,Binary}Section.Section`
- [x] Add support for `LITERAL+`
- [x] ~~Add IMAP4rev1 fallback for IDLE~~ (#543)
- [x] Consider restarting IDLE every 30min (#542)
- [x] Consider setting SEARCH charset for IMAP4rev1
- [ ] `ENABLE IMAP4rev2` on servers which also support IMAP4rev1
- [x] ~~Consider turning `SearchCriteria.Not` into a pointer, instead of a slice~~ (#505)
- [x] Add support for NAMESPACE
- [x] Add support for ESEARCH
- [x] Add support for SEARCHRES
- [x] Decode `literal8`
- [x] Make command methods extensible (see [RFC 4466](https://www.rfc-editor.org/rfc/rfc4466))
- [x] Canonicalize message flags and mailbox attrs
- [x] Parse date in envelope
- [x] Add tests which can run against any IMAP server
- [x] Consider turning lists of items into bool fields (e.g. `SearchOptions.Return`)
- [x] Drop `Client.StartTLS` in favor of a function which constructs an already-secured TLS connection
- [ ] Ignore responses that don't make sense given the connection state ([section 11.3](https://www.rfc-editor.org/rfc/rfc9051.html#section-11.3))
- [x] Consider re-using the same `*FetchBodySection` pointers in the returned map, or make it easier to match a `*FetchBodySection` from a command
- [ ] Drop `Command` because it's not extensible
- [ ] Some values can be either seq num or UID, how to represent these? Examples: `SearchData.Min/Max`, `SortCommand.Wait`, `ThreadData.Chain`. Should we add separate UID types for each?
- [ ] Some commands may return partial data, should we use pointers? See `StatusData`.
- [x] Add support for more extensions
  - [x] METADATA
  - [x] QUOTA (#492)
  - [x] SORT
  - [x] THREAD

PR #734, opened 2025-12-09
imapclient: Add a convenience function to enable whatever go-imap supports

arnt:
This allows a client to trust go-imap's judgment and not clutter its source with IMAP details.

### DIFF
--- a/imapclient/client.go
+++ b/imapclient/client.go
@@ -154,16 +154,17 @@ type Client struct {
 	decCh  chan struct{}
 	decErr error
 
-	mutex        sync.Mutex
-	state        imap.ConnState
-	caps         imap.CapSet
-	enabled      imap.CapSet
-	pendingCapCh chan struct{}
-	mailbox      *SelectedMailbox
-	cmdTag       uint64
-	pendingCmds  []command
-	contReqs     []continuationRequest
-	closed       bool
+	mutex           sync.Mutex
+	state           imap.ConnState
+	caps            imap.CapSet
+	enabled         imap.CapSet
+	enableAttempted bool
+	pendingCapCh    chan struct{}
+	mailbox         *SelectedMailbox
+	cmdTag          uint64
+	pendingCmds     []command
+	contReqs        []continuationRequest
+	closed          bool
 }
 
 // New creates a new IMAP client.
@@ -530,6 +531,7 @@ func (c *Client) completeCommand(cmd command, err error) {
 			c.state = imap.ConnStateNotAuthenticated
 			c.mailbox = nil
 			c.enabled = make(imap.CapSet)
+			c.enableAttempted = false
 			c.mutex.Unlock()
 		}
 	case *SelectCommand:
--- a/imapclient/client.go
+++ b/imapclient/client.go
@@ -438,7 +438,7 @@ func (c *Client) beginCommand(name string, cmd command) *commandEncoder {
 	}
 
 	c.pendingCmds = append(c.pendingCmds, cmd)
-	quotedUTF8 := c.caps.Has(imap.CapIMAP4rev2) || c.enabled.Has(imap.CapUTF8Accept)
+	quotedUTF8 := c.enabled.Has(imap.CapIMAP4rev2) || c.enabled.Has(imap.CapUTF8Accept)
 	literalMinus := c.caps.Has(imap.CapLiteralMinus)
 	literalPlus := c.caps.Has(imap.CapLiteralPlus)
 
@@ -533,6 +533,7 @@ func (c *Client) completeCommand(cmd command, err error) {
 			c.enabled = make(imap.CapSet)
 			c.enableAttempted = false
 			c.mutex.Unlock()
+			c.dec.QuotedUTF8 = false
 		}
 	case *SelectCommand:
 		if err == nil {
--- a/imapclient/enable.go
+++ b/imapclient/enable.go
@@ -74,7 +74,12 @@ func (c *Client) handleEnabled() error {
 	for name := range caps {
 		c.enabled[name] = struct{}{}
 	}
+	quotedUTF8 := c.enabled.Has(imap.CapIMAP4rev2) || c.enabled.Has(imap.CapUTF8Accept)
 	c.mutex.Unlock()
+
+	if quotedUTF8 {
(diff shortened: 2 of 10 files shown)


arnt:
I woke up this morning with mixed feelings.

This is a protocol, in the old Microsoft sense of the word: A series of function calls that don't necessarily make overwhelming sense to the user. [This](https://pkg.go.dev/github.com/emersion/go-imap/v2/imapclient) mentions 9051 and not 3501, but the code uses 3501 by default and (this is a three-step protocol description ⇒) to use 9051, you dial, login and then call EnableAll. Why that last? It makes sense to people with deep IMAP expertise, less sense to a more general audience. A typical "protocol" sequence of calls.

I'd like to do two things:
- make EnableAll the default,
- in a separate PR, add the remaining bits of support for [`UIDONLY`](https://www.rfc-editor.org/rfc/rfc9586.html) and change [FetchMessageBuffer](https://pkg.go.dev/github.com/emersion/go-imap/v2/imapclient#FetchMessageBuffer) to make `UID` mandatory and deprecate `SeqNum`.

This would simplify the API, make all client applications use the same set of extensions, and eliminate the possibility of bugs such as happen when people have the same SeqNum for different messages.

What do you think?

emersion:
> make EnableAll the default

I completely agree! (This was part of https://github.com/emersion/go-imap/issues/491)

We should probably provide a way to override this - e.g. with a customizable list of extensions to enable on connect in `Options`, which defaults to IMAP4rev2 (and friends).

> in a separate PR, add the remaining bits of support for [UIDONLY](https://www.rfc-editor.org/rfc/rfc9586.html) and change [FetchMessageBuffer](https://pkg.go.dev/github.com/emersion/go-imap/v2/imapclient#FetchMessageBuffer) to make UID mandatory and deprecate SeqNum

FWIW, we already ensure UID is always populated for UID FETCH:
https://github.com/emersion/go-imap/blob/7ac47a9cfd9a06ff5f3bdee52d7420137f023a1e/imapclient/fetch.go#L46

I don't know if we should be enabling `UIDONLY` by default, this sounds like a pretty large change. But maybe we should aim for that. Will think about this some more.

aleksandr4842:
Enable `UIDONLY` by default - sounds so good

arnt:
It's a digression, perhaps, but on the subject of UIDONLY: Enabling `UIDONLY` by default and removing `SeqNum`  has three effects:
1. It permits the occasional speedup (this is not a big factor at the moment)
2. It prevents client programmers from using MSN arithmetic (extremely rare)
3. It prevents client programmers from confusing UIDs and MSNs, one of the [most common problems on Stack Overflow](https://stackoverflow.com/questions/79835866/).

IMO ③ outweighs ② by a very large factor, and ① is potentially helpful (SELECTing a large folder via IMAP is none too fast).

emersion:
To somewhat mitigate ③, go-imap has a separate `imap.UID` type for UIDs.

While we're talking about seqnums, I had a question for a while: how does one implement paging without seqnums? The alps webmail needs to display pages of messages. Each page needs to display a fixed number of messages. This is straightforward using seqnums. How does one implement the same functionality with `UIDONLY`?

foxcpp:
Indeed, IMAP implements paging via sequence numbers, removing them will require the client to maintain full local message list and do paging locally. Sequence numbers allow storing very little data on the client - UI operations directly correspond to IMAP commands.

arnt:
Using sequence numbers was the recommended way to implement paging, ie. it was how Mark Crispin intended it to work. I know quite a few clients that did it, including the one I mostly use. However, I also know quite a few clients that stopped doing it, because it causes a poor user experience when you combine it with threading, and people seem to prefer threading over paging when given the choice.

If you ask for the 20 latest messages and combine messages into threads, sometimes you display 20 threads and sometimes you display just two threads.

There are exceptions — as I mentioned, I use a client that uses paging.

This isn't the area I focus on when I analyse IMAP clients, but I don't look away eityher. I don't think there's one single dominant solution. I've seen `UID THREAD … OR UNSEEN SINCE …` for example, which can be used to give you all threads that contain either new or unread messages.

emersion:
Some clients might prefer to display individual messages, and some clients might prefer to show message threads. I don't think go-imap should dictate which UI clients should present to their users.

arnt:
@emersion your code, your call, of course.

My distaste for MSNs may be coloured by my years of answering [questions where people get this wrong](https://stackoverflow.com/search?q=%5Bimap%5D+search+all). MSN arithmetic can be used to paginate messages, but AFAICT most people who paginate messages do it by issuing something like `SEARCH ALL` or `UID SEARCH ALL` and paginating in the client, which even makes sense in many clients. Sigh.

arnt:
@emersion I think you may find a message from me in your spam folder.

arnt:
Hi. Any comments on this? It's magnificently simple.

emersion:
I would prefer this to be the default, rather than an opt-in.

arnt:
Makes sense. I can update the PR to silently enable everything unless the client explicitly chooses to enable a (perhaps empty) set of extensions.

In that case, how do you prefer to handle the publicly available MSNs? If you ENABLE UIDONLY, you lose MSNs. I see two options:
1. Remove the MSNs from go-imap's API.
2. Document that MSNs are all 0 by default, but leave them in the API for clients that explicitly choose not to enable the relevant extension.

I don't see a lot of clients *use* MSNs except when forced to, but they can be useful in principle, so it's a choice.

arnt:
I pushed a new version now:
- if a user calls Enable(), the user gets those extensions
- if not, calling Select()/Examine() enables a default set

This should lead to most clients using the same set of extensions.

While doing that, I discovered some bugs, fixed them, and added some tests.

The IMAP4rev2 status is tricky. imapclient can send rev2 syntax if .Caps().Has(CapIMAP4rev2), but if it wants the server to send rev2 syntax, it has to enable rev2 explicitly. Therefore, Enable(...) ought to be called before any command that expects/needs rev2 responses. Tricky.

arnt:
The last force-push just changed indentation to suit the CI. The code still enables everything automatically by default, as [requested](https://github.com/emersion/go-imap/pull/734#issuecomment-4215982948) and unless the application developer chooses a different set of extensions explicitly.

PR #743, opened 2026-03-01
imapclient: handle unsolicited LIST responses in UnilateralDataHandler

WhyNotHugo:
When subscribing to MailboxName or SubscriptionChange via the NOTIFY extension, the server may send unsolicited LIST responses. Handle these via UnilateralDataHandler.

All other responses which can be triggered via NOTIFY are already properly handled.


emersion (approved):
LGTM, thanks!

PR #746, opened 2026-03-15
imapserver: Check nil pointer before use value

aleksandr4842:


emersion:
What is the motivation for this?

If the client has requested an item, why would the server not send it?

aleksandr4842:
When I implement IMAPRev1, client send recent option in status, but server don't add this and as result panic

aleksandr4842:
If panic is ok for this case -- I close PR

aleksandr4842:
I can write tests for show what I mean

aleksandr4842:
@emersion I rework this PR.

What about "If the client has requested an item, why would the server not send it?" -- server may not send if dev doesn't do it

Panic isn't good way

emersion:
If the server doesn't return an item the client asked for, I'd say it's a server implementation bug. It breaks the backend API contract. I'd say a panic is fine in this situation.

aleksandr4842:
Thank you

PR #748, opened 2026-03-24
imapclient: wrap inner errors

WhyNotHugo:
The following error (split here into multiple lines for readability):

    in response: cannot read tag:
    read tcp 192.168.2.3:51198->185.97.174.199:993:
    read: connection reset by peer

Does not match the following condition:

    errors.Is(err, syscall.ECONNRESET)

Calls to fmt.Errorf use `%v`, which embeds the inner error's string representation, without wrapping the inner error.

Use `%w` instead, to wrap the inner error, enabling usages such as the example above.

### DIFF
--- a/imapclient/acl.go
+++ b/imapclient/acl.go
@@ -64,7 +64,7 @@ func (cmd *GetACLCommand) Wait() (*GetACLData, error) {
 func (c *Client) handleMyRights() error {
 	data, err := readMyRights(c.dec)
 	if err != nil {
-		return fmt.Errorf("in myrights-response: %v", err)
+		return fmt.Errorf("in myrights-response: %w", err)
 	}
 	if cmd := findPendingCmdByType[*MyRightsCommand](c); cmd != nil {
 		cmd.data = *data
@@ -75,7 +75,7 @@ func (c *Client) handleMyRights() error {
 func (c *Client) handleGetACL() error {
 	data, err := readGetACL(c.dec)
 	if err != nil {
-		return fmt.Errorf("in getacl-response: %v", err)
+		return fmt.Errorf("in getacl-response: %w", err)
 	}
 	if cmd := findPendingCmdByType[*GetACLCommand](c); cmd != nil {
 		cmd.data = *data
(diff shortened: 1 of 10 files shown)


emersion:
In principle, sounds fine to me, but see the discussion in https://github.com/emersion/go-imap/pull/685

WhyNotHugo:
I hadn't thought of the fact that the become part of the API.

I believe that `StatusResponse` is the only public error declared in this library, and exposing that one sounds perfectly reasonable.

Other errors all seem to come from the stdlib and are quite likely to remain stable. E.g. `Decoder.Error` gets values such as `io.ErrUnexpectedEOF` which should not change.

I wanted to keep my changes focused, but simply wrapping the error in `in response: cannot read tag: %w` is insufficient: the error assigned to `Decoder.err` also wraps another error, and I think a large amount of code paths can yield `syscall.ECONNRESET`.

emersion:
> Other errors all seem to come from the stdlib and are quite likely to remain stable

The main problem isn't that errors coming from functions we're using might change, it's that changing our code might result in different errors being generated and break users. Effectively `%w` is an API promise.

internal/internal.go
emersion, line 1: Can we leave this file unchanged? I'd prefer not to allow API users to special-case `time.Parse` errors.

internal/imapwire/decoder.go
```diff
@@ -72,7 +72,7 @@ func NewDecoder(r *bufio.Reader, side ConnSide) *Decoder {
 
 func (dec *Decoder) mustUnreadByte() {
 	if err := dec.r.UnreadByte(); err != nil {
-		panic(fmt.Errorf("imapwire: failed to unread byte: %v", err))
+		panic(fmt.Errorf("imapwire: failed to unread byte: %w", err))
```
emersion: This is a panic, so I don't think wrapping is a good idea here.

emersion (commented):
Other than that, looks good to me.

internal/imapwire/decoder.go
WhyNotHugo: Done.

internal/internal.go
WhyNotHugo: Done.

WhyNotHugo (commented):
All updated.

emersion (approved):
Thanks!

PR #751, opened 2026-04-06
Add is session conn over flag

aleksandr4842:
Need know when session is over in Close command for dec session metrics

### DIFF
--- a/imapserver/conn.go
+++ b/imapserver/conn.go
@@ -46,6 +46,8 @@ type Conn struct {
 
 	state   imap.ConnState
 	session Session
+
+	isSessionOver bool
 }
 
 func newConn(c net.Conn, server *Server) *Conn {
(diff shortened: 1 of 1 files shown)


aleksandr4842:
@emersion hi! I wrote to emersion channel in IRC about it, but i can't see messages history... Please review it and feel free if not ok and you may offer alternative way

emersion:
I don't understand. The session is created in `Conn.serve`, and `session.Close` is called when the function returns. In which situation would this flag be false when `session.Close` is called?

aleksandr4842:
@emersion i'm so sorry, I thought that session.Close use when client send CLOSE command. Thank you!

aleksandr4842:
Please forgive me for push

emersion:
No worries

PR #755, opened 2026-05-16
imapclient: don't tear down the connection on dynamic COPYUID

### LINKED ISSUE #749: Malformed COPYUID response with wildcard causes fatal connection teardown
## Summary

When a server returns a COPYUID response code containing `*` (e.g., `OK [COPYUID 1234567890 1:5 100:*]`), the `readRespCodeCopyUID` function in `copy.go` returns an error ("dynamic number set in COPYUID response"). Because this parsing happens in the background read goroutine, the error propagates up through `readResponse()` → `read()`, which sets `decErr`, calls `closeWithError`, completes all pending commands with the error, and closes the TCP connection.

## Problem

The server's COPYUID response violates RFC 4315 (which forbids `*` in COPYUID UID sets), but the library's response — tearing down the entire connection — is disproportionate. The MOVE/COPY operation itself succeeded on the server; only the response metadata is malformed.

This means:
- The MOVE command doesn't just fail — every pending command on the connection fails
- The connection is permanently dead and must be re-established
- The client has no way to gracefully degrade (e.g., ignore the UID mapping and re-sync)

## Expected behavior

A malformed COPYUID response should be treated as a recoverable condition:
- Return `nil` for the UID mapping fields in `MoveData`/`CopyData` (as if the server didn't send COPYUID at all)
- Log or surface a warning
- Allow the connection to continue operating

This matches the behavior when a server simply doesn't support UIDPLUS — the client gets no UID mapping but the connection stays alive.

## Reproduction

This occurs in practice with Purelymail's IMAP server, which sends `*` in COPYUID responses to MOVE commands. We haven't tested with other providers.

## Affected code

`readRespCodeCopyUID()` in `copy.go`:

```go
if srcUIDs.Dynamic() || dstUIDs.Dynamic() {
    return 0, nil, nil, fmt.Errorf("imapclient: server returned dynamic number set in COPYUID response")
}
```

Returning an error here propagates to the read goroutine and kills the connection. Returning `(0, nil, nil, nil)` or equivalent would allow graceful degradation.

## Version

v2.0.0-beta.5

c-tonneslan:
Closes #749.

\`readRespCodeCopyUID\` returns an error when the server's COPYUID set contains \`*\`. RFC 4315 forbids that, but some real servers (Purelymail per the issue) send it anyway after MOVE. The error then bubbles up through the background read loop, sets \`decErr\`, calls \`closeWithError\`, and tears down the connection — so a single malformed advisory field fails every pending command on the connection and forces a reconnect, even though the MOVE itself already succeeded on the server.

Match the no-UIDPLUS path instead: when the response is dynamic, drop the UID mapping (zero \`uidValidity\`, nil source/dest sets) and return no error. The caller sees what it would see from a server that doesn't support UIDPLUS at all — empty \`CopyData.SourceUIDs\` / \`DestUIDs\` — and the connection stays up.

Tested locally with \`go test ./imapclient/...\`, all green.

### DIFF
--- a/imapclient/copy.go
+++ b/imapclient/copy.go
@@ -1,8 +1,6 @@
 package imapclient
 
 import (
-	"fmt"
-
 	"github.com/emersion/go-imap/v2"
 	"github.com/emersion/go-imap/v2/internal/imapwire"
 )
@@ -31,7 +29,14 @@ func readRespCodeCopyUID(dec *imapwire.Decoder) (uidValidity uint32, srcUIDs, ds
 		return 0, nil, nil, dec.Err()
 	}
 	if srcUIDs.Dynamic() || dstUIDs.Dynamic() {
-		return 0, nil, nil, fmt.Errorf("imapclient: server returned dynamic number set in COPYUID response")
+		// RFC 4315 forbids "*" in COPYUID UID sets, but some servers
+		// (Purelymail observed in #749) send it anyway after MOVE. The
+		// COPY/MOVE itself already succeeded on the server side; only
+		// the response metadata is malformed. Treat it the same as a
+		// server that doesn't support UIDPLUS at all — return empty UID
+		// mapping and no error — so the read loop doesn't tear down the
+		// whole connection over an advisory field.
+		return 0, nil, nil, nil
 	}
 	return uidValidity, srcUIDs, dstUIDs, nil
 }


arnt:
This made me wonder about a more general question — is there ever a response code that cannot safely be ignored? I can't think of one. In `* OK [FOO] blah blah`, FOO is a response code rather than a separate response precisely because it can be ignored, I think. It's additional information, not critical information.

So this PR is good, and perhaps other response code handling should be changed in the same way.

emersion (approved):
Thanks!

emersion:
> other response code handling should be changed in the same way

I'd prefer to keep the parser as strict as possible, to reduce risk of getting more non-conformant implementations.

PR #757, opened 2026-06-20
imapclient: support QRESYNC

hstern:
Adds client support for the QRESYNC `VANISHED` response (RFC 7162).

- Parses `* VANISHED [(EARLIER)] <uid-set>` and delivers the expunged UIDs to a new `UnilateralDataHandler.Vanished(uids imap.UIDSet, earlier bool)` callback, mirroring how `EXPUNGE` is routed.
- `FetchOptions.Vanished` adds the `VANISHED` modifier to a UID `FETCH` that carries `CHANGEDSINCE` — `(CHANGEDSINCE <modseq> VANISHED)` — so a client can fetch the changed messages and learn which UIDs were expunged since a mod-sequence in a single round trip.

This is the piece needed for incremental resync — detecting expunges without re-listing the whole mailbox. It covers the FETCH and unsolicited paths (any `VANISHED` response reaches the handler); the `SELECT (QRESYNC ...)` parameter is intentionally left out of this PR (happy to follow up). The server must have QRESYNC enabled via `client.Enable(imap.CapQResync)`.

Tests use a raw `net.Pipe` server since `imapmemserver` doesn't implement QRESYNC: one asserts a `VANISHED (EARLIER) 1:3,7` response reaches the handler with the right UIDs and `earlier` flag; the other asserts the FETCH modifier is emitted as `(CHANGEDSINCE 100 VANISHED)`.

Context: I'm building a Microsoft Graph mailbox server that maps `/me/messages/delta` onto IMAP CONDSTORE. `CHANGEDSINCE` already gives new + flag-changed messages; this lets the same delta also report deletions.

Developed with [Claude Code](https://claude.com/claude-code).

### DIFF
--- a/imapclient/client.go
+++ b/imapclient/client.go
@@ -986,6 +986,8 @@ func (c *Client) readResponseData(typ string) error {
 		return c.handleFetch(num)
 	case "EXPUNGE":
 		return c.handleExpunge(num)
+	case "VANISHED":
+		return c.handleVanished()
 	case "SEARCH":
 		return c.handleSearch()
 	case "ESEARCH":
@@ -1201,6 +1203,13 @@ type UnilateralDataHandler struct {
 	Expunge func(seqNum uint32)
 	Mailbox func(data *UnilateralDataMailbox)
 	Fetch   func(msg *FetchMessageData)
+
+	// Called when the server sends a VANISHED response (RFC 7162), reporting the
+	// UIDs of expunged messages. earlier is true for VANISHED (EARLIER), sent in
+	// response to a QRESYNC SELECT or a FETCH with the VANISHED modifier for
+	// messages expunged earlier; false for a VANISHED reporting a live expunge.
+	// Requires QRESYNC.
+	Vanished func(uids imap.UIDSet, earlier bool)
 
 	// Requires ENABLE METADATA or ENABLE SERVER-METADATA.
 	Metadata func(mailbox string, entries []string)
--- a/imapclient/client.go
+++ b/imapclient/client.go
@@ -1204,10 +1204,12 @@ type UnilateralDataHandler struct {
 	Mailbox func(data *UnilateralDataMailbox)
 	Fetch   func(msg *FetchMessageData)
 
-	// Called when the server sends a VANISHED response (RFC 7162), reporting the
-	// UIDs of expunged messages. earlier is true for VANISHED (EARLIER), sent in
-	// response to a QRESYNC SELECT or a FETCH with the VANISHED modifier for
-	// messages expunged earlier; false for a VANISHED reporting a live expunge.
+	// Called when the server sends an unsolicited VANISHED response (RFC 7162),
+	// reporting the UIDs of expunged messages. earlier indicates whether the
+	// response covers messages that were expunged earlier (VANISHED (EARLIER),
+	// sent in reply to a QRESYNC SELECT or a FETCH with the VANISHED modifier)
+	// rather than a live expunge.
+	//
 	// Requires QRESYNC.
 	Vanished func(uids imap.UIDSet, earlier bool)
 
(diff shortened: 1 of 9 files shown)


emersion:
This seems like a duplicate of https://github.com/emersion/go-imap/pull/720

hstern:
Thanks for the pointer — and apologies for the overlap with #720.

I've pushed an update that turns this PR into a superset of both branches rather than a parallel one. It now also includes the `SELECT (QRESYNC …)` resynchronization that #720 adds (`SelectOptions.QResync`, `SelectData.VanishedUIDs`, and routing of an in-flight `VANISHED` response onto the SELECT command), on top of the `FETCH … (CHANGEDSINCE n VANISHED)` modifier and the `UnilateralDataHandler.Vanished` callback that were already here.

Compared to #720, the main differences are:

- **Tests** — `TestSelectQResync` covers the `(QRESYNC (uidvalidity modseq known-uids (seq uid)))` encoding and the SELECT-time VANISHED routing; `TestVanished*` cover the FETCH modifier and the parsing edge cases. #720 has none.
- **VANISHED modifier parsing** is case-insensitive and tolerates an unrecognized `(modifier)` token instead of erroring — the remainder is always `SP known-uids`, so the response stays recoverable, matching how unknown `resp-text-code`s are discarded elsewhere in the client. It still errors on structural breakage.

Credit to @WhyNotHugo for the `SELECT (QRESYNC …)` shape in #720 — happy to reconcile however you prefer: land this as the combined PR, or I can split out the SELECT piece / rebase onto #720, whichever keeps the history cleanest for you.

imapclient/client.go
```diff
@@ -1202,6 +1204,13 @@ type UnilateralDataHandler struct {
 	Mailbox func(data *UnilateralDataMailbox)
 	Fetch   func(msg *FetchMessageData)
 
+	// Called when the server sends a VANISHED response (RFC 7162), reporting the
```
WhyNotHugo: `when the server sends an untagged VANISHED response`

(Clarify "untagged" here)

imapclient/client.go
```diff
@@ -1202,6 +1204,13 @@ type UnilateralDataHandler struct {
...
+	// UIDs of expunged messages. earlier is true for VANISHED (EARLIER), sent in
```
WhyNotHugo: > earlier is true for VANISHED (EARLIER)

This sentence doesn't make sense. Perhaps "earlier indicates whether this response covers earlier expunges".

imapclient/expunge.go
```diff
@@ -22,6 +24,45 @@ func (c *Client) UIDExpunge(uids imap.UIDSet) *ExpungeCommand {
...
+	if cmd := findPendingCmdByType[*SelectCommand](c); cmd != nil {
+		cmd.data.VanishedUIDs = append(cmd.data.VanishedUIDs, uids...)
+		return nil
+	}
+	if handler := c.options.unilateralDataHandler().Vanished; handler != nil {
+		handler(uids, earlier)
+	}
+	return nil
+}
+
```
WhyNotHugo: I guess this might be a matter of taste, but I'd put this in a separate `vanish.go` file rather than here; I don't see how this relates to the EXPUNGE command.

WhyNotHugo (commented):
I think you can rebase on top of #720 for simplicity, but honestly any approach is fine by me.

Appreciate the additional tests for the VANISHED bits.

hstern:
Thanks for the review @WhyNotHugo! Addressed all three in dabdeca:

- **`client.go` doc comments** — the `Vanished` handler comment now says "unsolicited VANISHED response" and describes `earlier` as *indicating whether the response covers messages expunged earlier* rather than the old "earlier is true for VANISHED (EARLIER)" phrasing.
- **`expunge.go`** — moved `handleVanished` into a dedicated `vanished.go` (pairing with the existing `vanished_test.go`), since it isn't related to the EXPUNGE command.

On the rebase-onto-#720 question: since #720 has no tests and this branch now carries both the SELECT resync and the FETCH `VANISHED` modifier plus coverage for them, I've kept it as the combined branch — but I'm still happy to rebase onto #720 or split out the SELECT piece if @emersion would prefer that for the history. Just let me know.

PR #758, opened 2026-06-23
Remove redundant setReadTimeout

aleksandr4842:
We set read timeout in 176 line. Removed setter rewrite our timeout

### DIFF
--- a/imapserver/conn.go
+++ b/imapserver/conn.go
@@ -183,7 +183,6 @@ func (c *Conn) serve() {
 			break
 		}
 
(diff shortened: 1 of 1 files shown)


imapserver/conn.go
emersion, line 182: The `dec.EOF()` call here blocks until it sees more input or EOF. The read timeout should be set to `idleReadTimeout` before the call if authenticated.

aleksandr4842: Thanks

PR #761, opened 2026-07-09
imapclient: send an empty SASL continuation as an empty line

### LINKED ISSUE #760: imapclient: zero-length SASL continuation response sent as `=` is rejected by strict servers
`imapclient`'s `AUTHENTICATE` handling encodes a zero-length SASL *continuation* response as a single `=`, which is not valid in that position. Servers that strictly validate the continuation as base64 (e.g. Dovecot) reject it.

Root cause — `imapclient/authenticate.go`, `writeSASLResp`:

```go
respStr := internal.EncodeSASL(resp)
```

`internal.EncodeSASL` returns `"="` when `resp` is empty. That is the RFC 4959 zero-length *initial-response* marker, valid only in the `AUTHENTICATE <mech> =` command line. For a continuation response (client data sent after a `+` challenge), RFC 3501 specifies base64; there is no `=` convention, and `=` alone is not valid base64. Dovecot reports `Invalid base64 data in continued response` and aborts the exchange.

This only affects mechanisms that emit a zero-length continuation token. Single-round mechanisms (PLAIN, OAUTHBEARER) never hit it, which is why it hasn't surfaced. It shows up with SASL GSSAPI (RFC 4752): after the server's AP-REP mutual-authentication token, `GSS_Init_sec_context` returns `GSS_S_COMPLETE` with a zero-length output token, which the client must send as an empty continuation before the server sends its security-layer offer.

Reproduction: authenticate against a Dovecot server configured with `auth_mechanisms = gssapi` using a SASL GSSAPI client. The context establishes (`security context state completed` in Dovecot's auth log), then the empty continuation is rejected with `Invalid base64 data in continued response`.

Expected: an empty continuation response is sent as an empty line (base64 of empty input is the empty string).
Actual: sent as `=`, which the server rejects.

Fix: in `writeSASLResp`, base64-encode the response directly rather than via `EncodeSASL`, so an empty response encodes to `""`. `EncodeSASL` (with the `=` rule) remains correct for the initial-response slot in `Authenticate`.

hstern:
Fixes #760.

`imapclient.writeSASLResp` encodes a zero-length continuation response as `=` (via `internal.EncodeSASL`). `=` is the RFC 4959 zero-length *initial-response* marker, valid only in the `AUTHENTICATE <mech> =` command line, not in a continuation, where it is not valid base64. Strict servers such as Dovecot reject it (`Invalid base64 data in continued response`).

This surfaces with multi-step mechanisms whose client emits a zero-length continuation token — e.g. SASL GSSAPI (RFC 4752), whose mutual-authentication step sends an empty token after the server's AP-REP. Single-round mechanisms (PLAIN, OAUTHBEARER) are unaffected.

The fix base64-encodes the continuation directly, so an empty response becomes an empty line. `EncodeSASL` (with the `=` rule) is unchanged and still used for the initial-response slot in `Authenticate`.

### DIFF
--- a/imapclient/authenticate.go
+++ b/imapclient/authenticate.go
@@ -1,6 +1,7 @@
 package imapclient
 
 import (
+	"encoding/base64"
 	"fmt"
 
 	"github.com/emersion/go-sasl"
@@ -76,7 +77,12 @@ type authenticateCommand struct {
 }
 
 func (c *Client) writeSASLResp(resp []byte) error {
-	respStr := internal.EncodeSASL(resp)
+	// A continuation response is base64 (RFC 3501). A zero-length response must be
+	// an empty line: the "=" zero-length marker (RFC 4959) is valid only in the
+	// AUTHENTICATE initial-response slot, not in a continuation, where strict
+	// servers (e.g. Dovecot) reject "=" as invalid base64. This matters for
+	// multi-step mechanisms like GSSAPI, whose mutual-auth step sends an empty token.
+	respStr := base64.StdEncoding.EncodeToString(resp)
 	if _, err := c.bw.WriteString(respStr + "\r\n"); err != nil {
 		return err
 	}


imapclient/authenticate.go
```diff
@@ -76,7 +77,12 @@ type authenticateCommand struct {
 }
 
 func (c *Client) writeSASLResp(resp []byte) error {
-	respStr := internal.EncodeSASL(resp)
```
emersion, line 79: Can we rename `EncodeSASL` and `DecodeSASL` to `EncodeSASLInitialResp` and `DecodeSASLInitialResp`, to make it clear these are only used for initial response?

emersion (commented):
Note, I think the same issue affects the server part.

PR #765, opened 2026-09-05
imapserver: answer BAD for a parse failure, not NO [SERVERBUG]

0kaba0hub:
A malformed command is answered `NO [SERVERBUG] Internal server error` rather
than `BAD`:

```
a1 SEARCH HEADER Subject "fileinto test"   -> OK
a1 SEARCH HEADER Subject fileinto test     -> NO [SERVERBUG] Internal server error
```

The second is the client's syntax: an unquoted argument containing a space, so
`test` is read as the next search key. The server logs
`handling SEARCH command: in search-key: imap: bad number set value "TEST"`.

The connection loop already tells the cases apart by error type — an
`*imap.Error` is sent as the status response it carries, a
`*imapwire.DecoderExpectError` becomes `BAD [CLIENTBUG]` with the syntax text,
anything else is logged and answered `NO [SERVERBUG]`. Three places returned a
plain error from a failed parse and so fell into the last branch:

- `Decoder.ExpectNumSet`, where `imapnum.ParseSet` refuses the sequence-set,
  and `Decoder.ExpectMailbox`, where the name is not modified UTF-7: both are
  expect methods and now return a `DecoderExpectError` like the others;
- `readSearchKeyWithAtom`'s default branch, which parses an unknown search key
  as a sequence-set candidate — the one clients actually hit. It is not an
  expect method, so it returns an `imap.Error` with `BAD` and `CLIENTBUG`.

Left alone deliberately: the literal reader's copy failure. A client that
disappears mid-literal is a connection error rather than a syntax one, and
classifying it as either without deciding the right answer seemed wrong — happy
to follow up if you have a preference.

Tests check the error type each function returns; the mapping from error type
to status line is the connection loop's existing behaviour.

### DIFF
--- a/imapserver/parse_error_wire_test.go
+++ b/imapserver/parse_error_wire_test.go
@@ -1,0 +1,128 @@
+package imapserver_test
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"net"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/emersion/go-imap/v2"
+	"github.com/emersion/go-imap/v2/imapserver"
+	"github.com/emersion/go-imap/v2/imapserver/imapmemserver"
+)
+
+// parseWire is a raw connection to a server: these cases are about the exact
+// status line, which a client library hides.
+type parseWire struct {
+	t    *testing.T
+	conn net.Conn
+	rd   *bufio.Reader
+}
+
+func dialParseServer(t *testing.T, newSession func(imapserver.Session) imapserver.Session) *parseWire {
+	t.Helper()
+	mem := imapmemserver.New()
+	u := imapmemserver.NewUser("u", "p")
+	if err := u.Create("INBOX", nil); err != nil {
+		t.Fatal(err)
+	}
+	mem.AddUser(u)
+	srv := imapserver.New(&imapserver.Options{
+		NewSession: func(*imapserver.Conn) (imapserver.Session, *imapserver.GreetingData, error) {
+			s := imapserver.Session(mem.NewSession())
+			if newSession != nil {
+				s = newSession(s)
+			}
+			return s, nil, nil
+		},
+		InsecureAuth: true,
+		Caps:         imap.CapSet{imap.CapIMAP4rev1: struct{}{}},
+	})
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() { ln.Close() })
+	go srv.Serve(ln) //nolint:errcheck
+
+	c, err := net.Dial("tcp", ln.Addr().String())
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() { c.Close() })
+	w := &parseWire{t: t, conn: c, rd: bufio.NewReader(c)}
+	w.line()
+	fmt.Fprint(w.conn, "a1 LOGIN u p\r\n")
+	w.until("a1")
+	return w
+}
+
+func (w *parseWire) line() string {
+	w.t.Helper()
+	_ = w.conn.SetReadDeadline(time.Now().Add(3 * time.Second))
+	l, err := w.rd.ReadString('\n')
+	if err != nil {
+		w.t.Fatalf("read: %v", err)
+	}
+	return strings.TrimRight(l, "\r\n")
+}
+
+// until returns the tagged status line for tag.
+func (w *parseWire) until(tag string) string {
+	w.t.Helper()
+	for {
+		l := w.line()
+		if strings.HasPrefix(l, tag+" ") {
+			return l
+		}
+	}
+}
+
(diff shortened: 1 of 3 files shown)


imapserver/search.go
```diff
@@ -331,7 +331,11 @@ func readSearchKeyWithAtom(criteria *imap.SearchCriteria, dec *imapwire.Decoder,
 	default:
 		seqSet, err := imapwire.ParseSeqSet(key)
 		if err != nil {
-			return err
+			// An unknown key reaches here as a sequence-set candidate; failing
+			// to parse it is the client's syntax, not a server fault.
+			return &imapwire.DecoderExpectError{
```
emersion: `imapwire.DecoderExpectError` is specifically for decoder errors returned by "expect" style methods. This is not one of them.

`imap.Error` would be more appropriate here.

internal/imapwire/decoder.go
```diff
@@ -518,10 +518,15 @@ func (dec *Decoder) ExpectMailbox(ptr *string) bool {
 		return true
 	}
 	name, err := utf7.Decode(name)
-	if err == nil {
-		*ptr = name
+	if err != nil {
+		// A name that is not modified UTF-7 is a malformed command: the loop
+		// answers BAD for an expect error, NO [SERVERBUG] for anything else.
```
emersion: This refers to the server loop, but this is also called from the client. I'm not sure the comment is necessary here.

internal/imapwire/decoder.go
```diff
@@ -545,7 +550,10 @@ func (dec *Decoder) ExpectNumSet(kind NumKind, ptr *imap.NumSet) bool {
 	}
 	numSet, err := imapnum.ParseSet(s)
 	if err != nil {
-		return dec.returnErr(err)
+		// Same: the client sent what the grammar does not allow.
```
emersion: Ditto, and "same" is unclear here.

imapserver/parse_error_wire_test.go
```diff
@@ -0,0 +1,128 @@
...
+		},
+		InsecureAuth: true,
+		Caps:         imap.CapSet{imap.CapIMAP4rev1: struct{}{}},
+	})
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() { ln.Close() })
+	go srv.Serve(ln) //nolint:errcheck
```
emersion: Can we leave out `nolint` comments? This project doesn't use a linter.

imapserver/parse_error_wire_test.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// A plain error from the handler is still NO [SERVERBUG]: this classifies parse
+// failures, it does not answer BAD to everything.
+func TestHandlerErrorIsServerBug(t *testing.T) {
+	w := dialParseServer(t, func(s imapserver.Session) imapserver.Session {
+		return &failingSelect{Session: s}
+	})
+	fmt.Fprint(w.conn, "a2 SELECT INBOX\r\n")
+	if got := w.until("a2"); !strings.Contains(got, "SERVERBUG") {
```
emersion: It seems weird to check for `SERVERBUG` specifically here… Shouldn't we check that we got the error we expect instead?

imapserver/parse_error_wire_test.go
emersion: I'm not a huge fan of ad hoc string parsing in these tests.

imapserver/search.go
0kaba0hub: Right — this isn't an expect method. It now returns an `imap.Error` with `BAD` and `CLIENTBUG`, which the connection loop sends as-is. The two decoder sites stay `DecoderExpectError`, since those are expect methods.

internal/imapwire/decoder.go
0kaba0hub: Dropped. The decoder is shared with the client, so the comment didn't belong here; the error text says what happened.

internal/imapwire/decoder.go
0kaba0hub: Dropped as well.

imapserver/parse_error_wire_test.go
0kaba0hub: Removed; the test file it was in is gone.

imapserver/parse_error_wire_test.go
0kaba0hub: Agreed — that test asserted the loop's existing behaviour through a substring, which isn't what this change is about. Removed.

imapserver/parse_error_wire_test.go
0kaba0hub: Replaced the wire test with unit tests that check the error type each function returns: `readSearchKey` yields an `*imap.Error` with `BAD`/`CLIENTBUG`, `ExpectNumSet` and `ExpectMailbox` yield a `*DecoderExpectError`. The mapping from error type to status line is the connection loop's existing behaviour, so the tests stop there.

emersion (approved):
Thanks!

PR #768, opened 2026-09-15
imap: echo and read the FETCH partial offset as a 64-bit number

youdie006:
The same `imap.SectionPartial.Offset` is carried by four sites, and two of them are 32-bit:

| site | direction | width |
| --- | --- | --- |
| `imapclient/fetch.go:146` `writeSectionPartial` | client writes the request | `Number64` |
| `imapserver/fetch.go:319` | server decodes the request | `ExpectNumber64` |
| `imapserver/fetch.go:426` `writeItemBodySection` | server echoes it back | `Number(uint32(...))` |
| `imapclient/fetch.go:1275` `readPartialOffset` | client reads the echo | `ExpectNumber` into a `uint32` |

So an offset above 2^32 survives the request and is truncated in the response. Asking `imapmemserver` for `BODY[]<4294967301.10>` gets back `BODY[]<5>`.

That matters because `matchFetchItemBodySection` compares the echoed offset against the one the client asked for, and deliberately ignores the rest (`imapclient/fetch.go:254-261`):

```go
if cmd.Partial != nil && cmd.Partial.Offset != resp.Partial.Offset {
    return false
}
// Ignore Partial.Size and Peek: these are not echoed back by the server
```

With the offset truncated, the client cannot match this server's response to its own request, and `FindBodySection` returns nil.

RFC 9051 spells it `partial = "<" number64 "." nz-number64 ">"`, so a >32-bit offset is legal. `Number64` (`internal/imapwire/encoder.go:212`) and `ExpectNumber64` (`decoder.go:360`) already exist and the sibling writer at `imapclient/fetch.go:146` already uses them.

This is the follow-up I noted in #766 as a separate finding.

<details>
<summary>Verification</summary>

Go 1.26.3. `md5sum` of both changed sources on every row. `go test -count=1 -run TestFetchPartialOffsetEcho ./imapclient/`

| row | `imapserver/fetch.go` | `imapclient/fetch.go` | result |
| --- | --- | --- | --- |
| this PR | `d8b90ec1` | `95fc588f` | pass |
| both pristine | `7a72e2a5` | `fdc6b1ae` | `echoed partial offset = &{5 0}, want 4294967301` |
| revert the server echo only | `7a72e2a5` | `95fc588f` | `echoed partial offset = &{5 0}, want 4294967301` |
| revert the client decode only | `d8b90ec1` | `fdc6b1ae` | `in section-spec: imapwire: expected number, got ">"` |
| echo `Offset + 1` | `1a381eda` | `95fc588f` | `echoed partial offset = &{4294967302 0}, want 4294967301` |

The two middle rows are the N-1 partial fixes, and they fail for different reasons: reverting the server truncates, reverting the client cannot parse the 64-bit number the server now writes. Both sites are independently required.

Gates as `.build.yml` runs them: `gofmt -l` clean, `go build -race ./...` ok, `go test -race ./...` ok on all four packages. `go vet ./...` reports 8 lines in `imapserver/message.go`, byte-identical to pristine; that file is untouched here.

Not run: the `test-dovecot` job needs a live Dovecot (`GOIMAP_TEST_DOVECOT=1`). Local Go is 1.26.3 while `go.mod` says 1.18; the change uses nothing newer.

</details>

One neighbour worth flagging: #663 adds the same `Number(uint32(partial.Offset))` line to `WriteBinarySection`. I left that alone since it is your branch, but it would want `Number64` too, along with `matchFetchItemBinarySection` at `imapclient/fetch.go:265`, which currently says the partial is not echoed back.

Disclosure: this change was written with AI assistance (Claude). The measurements above were reproduced and checked by me before opening this PR.


emersion:
The ABNF accepts 64-bit numbers in the `FETCH` request, however the response uses regular 32-bit numbers:

```abnf
msg-att-static  = "ENVELOPE" SP envelope /
                  "INTERNALDATE" SP date-time /
                  "RFC822.SIZE" SP number64 /
                  "BODY" ["STRUCTURE"] SP body /
                  "BODY" section ["<" number ">"] SP nstring /
                  "BINARY" section-binary SP (nstring / literal8) /
                  "BINARY.SIZE" section-binary SP number /
                  "UID" SP uniqueid
                    ; MUST NOT change for a message
```

